PR #9, opened 2020-10-21
Initial frontend commit using yarn, and inclusion of testing-library module

### LINKED ISSUE #2: Set up React boilerplate
Task: get a basic "Hello, world!" level boilerplate running for React in TypeScript. 

If you could integrate a testing library in there as well, that would be great! From a quick Google search, [React Testing Library](https://testing-library.com/) seems to fit our use case but feel free to bring up other alternatives. I'll also suggest to have `yarn` set up instead of `npm` for our dependency management on both the front-end and back-end.

That should be good for this ticket, though we can add more libraries (e.g. Redux) later on when we find the need to do so.

richardyjtian:
Resolves #2


engineereng (approved):

andrewzulaybar (changes requested):
This looks like a good start! Do you think you'd be able to add support for TypeScript? I think that's the only thing left that we'd need :)

andrewzulaybar (approved):
Looks good to me! Feel free to merge it in :)

ericrkuo (approved):

PR #10, opened 2020-10-22
Set up Node.js/Express.js

### LINKED ISSUE #3: Set up Node.js/Express.js
Task: get a basic "Hello, world!" level boilerplate running for Node.js and Express.js in TypeScript.

If you could integrate a testing library that would be great. I'm leaning towards [Jest](https://jestjs.io/) cause it's a pretty popular choice, but if you have some other ones that you'd prefer please bring them up here! I also suggested in #2 to have `yarn` set up instead of `npm` on the front-end so it would be nice to have that on the back-end as well.

Feel free to make suggestions on things we should add at the start, but I think this should be good for now. We can add other libraries later on as we see fit.

ericrkuo:
Resolves #3

- Added typescript with express
- Added ts-jest for testing


andrewzulaybar (approved):
This looks great! I also really like using nodemon so thanks for adding that in :)

engineereng (approved):

PR #19, opened 2020-10-29
Added Endpoints for Google Places API for Search and Details

ld357:
The Yelp API is currently down so I didn't write endpoints for those! If it does end up getting fixed before Saturday, I'll try to add those in. I wrote tests that just print the results so you can see what it returns.

The two endpoints define a call for GooglePlacesSearch and GooglePlacesDetails (Search is needed to get a few of the fields as well as the placeID needed for the call to Details)

The format is slightly messy for the URLs in the GET requests because it wasn't allowing me to format it in a parameters object but if anyone has a clue on how to make it cleaner and if we end up keeping these calls, feel free to let me know! 

I just set a few of the fields that we'll most likely need to use. There is quite a number of discussions that we'll have to have regarding which API and what fields we want given our current contstraints. 

Please refer to this document for more information- [Google Places vs. Yelp APIs](https://docs.google.com/document/d/1VB1m5T601XZMaO-l-brTVzfyISdjWBng9krJK-UmBUw/edit?usp=sharing)

To test, you'll have to get your own key for now from Google Cloud and paste it into your .ENV with key "GOOGLE_PLACES_API_KEY="

### DIFF
--- a/backend/controllers/GooglePlacesAPI.ts
+++ b/backend/controllers/GooglePlacesAPI.ts
@@ -1,0 +1,64 @@
+import {AxiosError, AxiosResponse} from "axios";
+
+class GooglePlacesAPI {
+
+    constructor() {
+        require('dotenv').config();
+    }
+
+    getGooglePlacesSearch(data: string): Promise<AxiosResponse<JSON>> {
+        let dataURI: string = encodeURIComponent(data);
+        let axios = require('axios');
+
+        return axios.get('https://maps.googleapis.com/maps/api/place/findplacefromtext/json?'
+            + 'input=' + dataURI
+            + '&inputtype=textquery'
+            + '&fields=formatted_address,name,icon,place_id,types,photos'
+            + '&key=' + process.env.GOOGLE_PLACES_API_KEY
+        )
+            .then((response: AxiosResponse) => {
+
+                let result: JSON = response.data
+                return Promise.resolve(result);
+
+            })
+            .catch((err: AxiosError) => {
+
+                //Google API error message
+                console.log(err?.response?.data?.error)
+
+                //Axios entire error message
+                console.log(err)
+
+                return Promise.reject(err);
+            });
+    }
+
+    getGooglePlacesDetails(placeID: string): Promise<AxiosResponse<JSON>> {
+        let axios = require('axios');
+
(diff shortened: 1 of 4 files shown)


ericrkuo:
> The format is slightly messy for the URLs in the GET requests because it wasn't allowing me to format it in a parameters object but if anyone has a clue on how to make it cleaner and if we end up keeping these calls, feel free to let me know!

I think this should work: https://github.com/axios/axios#request-config, see `params`. Looks good though!
image

ericrkuo (approved):

backend/controllers/GooglePlacesAPI.ts
```diff
@@ -0,0 +1,64 @@
...
+
+class GooglePlacesAPI {
+
+    constructor() {
+        require('dotenv').config();
+    }
+
+    getGooglePlacesSearch(data: string): Promise<AxiosResponse<JSON>> {
+        let dataURI: string = encodeURIComponent(data);
+        let axios = require('axios');
```
andrewzulaybar, line 11: I think you should be able to add this in the import statement up top by changing it to:
```javascript
import axios, {AxiosError, AxiosResponse} from "axios";
```

backend/controllers/GooglePlacesAPI.ts
```diff
@@ -0,0 +1,64 @@
...
+            .then((response: AxiosResponse) => {
+
+                let result: JSON = response.data
+                return Promise.resolve(result);
+
+            })
+            .catch((err: AxiosError) => {
+
+                //Google API error message
+                console.log(err?.response?.data?.error)
```
andrewzulaybar, line 28: The optional chaining 🙌

backend/controllers/GooglePlacesAPI.ts
```diff
@@ -0,0 +1,64 @@
...
+
+                //Axios entire error message
+                console.log(err)
+
+                return Promise.reject(err);
+            });
+    }
+}
+
+module.exports.GooglePlacesAPI = GooglePlacesAPI;
```
andrewzulaybar, line 64: You should be able to change this to ES6 syntax with:
```javascript
export { GooglePlacesAPI };
```

andrewzulaybar (approved):
Sorry for the delay, just getting around to this now! I made some minor suggestions but overall this is great - thanks for getting this set up!

PR #20, opened 2020-10-30
Proxy test PR

ericrkuo:
Setting up a simple proxy to direct requests to backend. Not sure if we want to use this, was just playing around.

Summary:
Right now FE is in `localhost:3000` and BE is in `localhost:8000`. With the proxy, client requests in the FE can be formatted like below, and requests will be proxied to the backend respectively. I don't know too much about forward/reverse proxy's, but this was something I was playing around with and think it can be helpful.

I believe this is a forward proxy, but please clarify if I'm wrong
https://github.com/ubclaunchpad/foodies/blob/9f0f4a0b254501c511feb51411cf0c450b31b968/frontend/src/App.tsx#L27

This might be a better solution: https://create-react-app.dev/docs/proxying-api-requests-in-development/#configuring-the-proxy-manually

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 ## ignore database
 /tmp
+.idea
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -51,5 +51,6 @@
       "yarn lint",
       "yarn prettier"
     ]
-  }
+  },
+  "proxy": "http://localhost:8000"
 }


andrewzulaybar (approved):
Love this idea! Thanks so much for getting this set-up, I think this will be helpful and make sending requests a lot simpler on the front-end.

We can go with this approach for now and then adapt it if necessary :)

ericrkuo:
Sweet sounds good!

richardyjtian (approved):

PR #21, opened 2020-10-30
Add typescript-eslint, Prettier, and Husky git hooks

andrewzulaybar:
For both the front-end and back-end, I added:
- `typescript-eslint` for linting
- `Prettier` for code formatting
- pre-commit hooks that run the `yarn lint` and `yarn prettier` scripts on each commit

Feel free to add any eslint rules that you like. The only difference between the back-end and front-end right now is the usage of quotes (single for back-end, double for front-end).

@richardyjtian Also just a heads up, I deleted the `react-testing` library on the front-end after some discussion at the leads meeting. We probably shouldn't be spending our time making front-end tests that are brittle and subject to change often.


frontend/src/serviceWorker.js
```diff
@@ -1,141 +0,0 @@
-// This optional code is used to register a service worker.
```
andrewzulaybar, line 1: I thought about this, and we probably don't need service workers?

ericrkuo (approved):

richardyjtian (approved):

ericrkuo:
What do you think about this @andrewzulaybar ?
https://typedoc.org/

PR #22, opened 2020-10-31
Added endpoints for Google Maps for zoom, center, and markers

### LINKED ISSUE #13: Look into Google Maps API integration
Research the limitations of the Google Maps API, and create a small service (this could just be functions for now) on the back-end that can connect to the API endpoints. Some things that we'd like to be able to do include:
- Being able to embed a map on our site
- How we can programmatically control a map
- How pinning locations works

I'm thinking that it would be best to have these API calls stored on the back-end but it's possible that the front-end might be a better place for it. Please report your findings here!

Upon a quick Google search, I think these two links might be a good place to start:
- https://developers.google.com/maps/documentation/javascript/overview
- https://developers.google.com/maps/documentation/embed/get-started

richardyjtian:
The google-maps-react module was used to easily render a map by providing data for the "zoom" field and longitude and latitude data for the "initialCenter" field. Markers can be displayed by giving each marker a longitude and latitude as well. As this is the case, it's best to have the API call in the frontend, but the data being programmatically decided by the backend.

As with the Google Places API, a Google Cloud key is needed to test with the Google Maps API, and this key should be placed in a .env file in the src folder with the line:
REACT_APP_GOOGLE_MAPS_API_KEY=

Resolves #13

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -13,6 +13,7 @@
 
 # misc
 .DS_Store
+.env
 .env.local
 .env.development.local
 .env.test.local
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -10,6 +10,7 @@
     "@types/node": "^14.14.3",
     "@types/react": "^16.9.53",
     "@types/react-dom": "^16.9.8",
+    "google-maps-react": "^2.0.6",
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
     "react-scripts": "3.4.4",
--- a/frontend/src/MapContainer.js
+++ b/frontend/src/MapContainer.js
@@ -1,0 +1,49 @@
+import React, { Component } from 'react';
+import { Map, GoogleApiWrapper, Marker} from 'google-maps-react';
+
+const mapStyles = {
+  width: '100%',
+  height: '100%',
+};
+
+class MapContainer extends Component {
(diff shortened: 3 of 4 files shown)


andrewzulaybar (approved):
Sorry for the delay! Thanks for looking into this, I like how this is looking 🙌

I was wondering, are you familiar with functional components? They're just an alternative to class components, so I was thinking moving forward we could look into styling our components like this because this is the direction that the React community is headed toward. Not a blocker for this PR though, but if you could change the MapContainer class to a functional component either this week or next that would be great! More info: https://reactjs.org/docs/components-and-props.html

richardyjtian:
@andrewzulaybar sorry I didn't see your comment but yes! I can definitely look into changing the map into a functional component!

richardyjtian:
I couldn't figure out how to change it to a functional component, but I did find this example that very much looks like what we're aiming for - http://google-map-react.github.io/google-map-react/map/main
It seems that most examples create the map as a class component

andrewzulaybar:
@richardyjtian I'm sorry I should have mentioned this earlier but I forgot about the comment I made 😅  I pulled your work from this branch into another branch last week and integrated it with the home page so it should all be good now. I haven't connected it with the promotions just yet though cause I'm waiting until #59 and #52 get merged in!

Issue #35, opened 2020-11-01
Add sample data for users


andrewzulaybar:
## Task 🔨
Create sample data according to the schema for the `User` table and the `Promotion` table. We can also add example queries for these tables so we have an idea of the queries we'll write in the controllers. I think @ericrkuo is also planning to get [`sequelize`](https://sequelize.org/) set up so that we don't have to deal with query strings, so maybe take a look at how you could make these queries using that library.

You can also create sample data for the other tables in the schema (`User_Uploads_Promotion`, `User_Saved_Promotion`, `Promotion_Has_Category`, `Promotion_Category`, `Promotion_Has_Discount`, `Discount`, and `CuisineType`) if you have time.

Database schema: https://docs.google.com/document/d/131JfzQXb472CyOL2yMSqbfuDyGI13UwQgnh0kowzPuk/edit

yunishimura0716:
I firstly made sample data for Users, restaurant, discount and promotion. I'll make by other data through the discussion with @ericrkuo

PR #43, opened 2020-11-03
Backend ORM - TypeORM

ericrkuo:
Issue: https://github.com/ubclaunchpad/foodies/issues/35

_Summary_: 
Setup backend with ORM using TypeORM, this would be an alternative to raw SQL. I talked with @andrewzulaybar about exploring ORMs for our backend project, and we could go with this approach or SQL approach depending on how everyone feels. Just a heads up, I don't know too much about databases, I haven't taken 304. A lot of my terminology and logic may be incorrect, please correct me if somethings off.

I think in order to understand this PR, it may be best to go through the TypeORM documentation first, otherwise, a lot of things won't make sense. I don't know a lot about ORMs so I had to learn most of this from scratch, their documentation is pretty good.  https://typeorm.io/#/. I would also recommend checking out this branch, installing Postgres, creating a database called foodies, run yarn install, and then debug the project to see how everything works.

_What is an ORM?_ 
- https://en.wikipedia.org/wiki/Object%E2%80%93relational_mapping

_Why an ORM?_
This library makes it very easy for us to manage database migrations, leverage transactions, insert/update/delete from the database, and more. As you can see, we have a majority of our backend database schema mostly finished with little code. That way we can focus more so on our business logic and not have to worry about our project getting really large.

_Short summary:_
- Entities
  - https://github.com/ubclaunchpad/foodies/tree/backend-orm/backend/src/data
  - this folder contains all our entities, each entity is translated into a table in Postgres. As you can see by the annotations, there is a lot of customizable behaviour with respect to unidirectional/bidirectional associations, constraints, enums
- Repositories
  - each repository allows us to insert/update/delete entities.
https://github.com/ubclaunchpad/foodies/blob/408e111a8016ea047eefeca02a6849e5b906af6d/backend/src/App.ts#L33-L35
- Establishing a connection with Postgres
  - all we need is this line as well as `ormconfig.json`
https://github.com/ubclaunchpad/foodies/blob/408e111a8016ea047eefeca02a6849e5b906af6d/backend/src/App.ts#L12

_Special features_
- eager loading vs lazy loading
  - this is a very useful feature if we want to enhance performance
  - See https://typeorm.io/#/eager-and-lazy-relations
https://github.com/ubclaunchpad/foodies/blob/408e111a8016ea047eefeca02a6849e5b906af6d/backend/src/App.ts#L48-L56
- we can always use raw SQL if there are queries too complicated for typeORM, see https://typeorm.io/#/entity-manager-api
- different patterns in terms of how we access data
  - https://typeorm.io/#/active-record-data-mapper
- supported query builder features
  - https://typeorm.io/#/select-query-builder

_Database schema changes_
- I made some changes to the DB schema, we can definitely revert back to what is said in the doc, but these are some things I changed.
1. Instead of `User_uploads_promotion` join table
  - we now have a bidirectional one to many relationship where `promotion` is the owning side (so promotion has a column called `useID` and user has an array of `uploadedPromotions`.
  - since it's bidirectional, we can now see which user uploaded a certain promotion, and we can also see all the promotions a user uploaded
2. instead of `Promotion_has_discount` join table
  - we now have a bidirectional one to many relationship where `discount` is the owning side (so discount has a column called `promotionId`
  - now we can access all the discounts related to a promotion, as well as all promotions related to a discount (for search query purposes)
  - also not a many to many, see google doc comment https://docs.google.com/document/d/131JfzQXb472CyOL2yMSqbfuDyGI13UwQgnh0kowzPuk/edit
3.  Changed `discount_type` to an enum
4. `promotion_category` and `promotion_has_category`
  - I made all these enums, I think an enum makes more sense since we would be defining all possible types of categories that the user can select from. And we don't have to create another separate table for this
5.  Changed `cuisine_type` to enum
  - this is one of the harder enums since there can be lots of cuisine types, but I think it makes more sense to have an enum. This makes it easier to validate whether this is a cuisine type (so we prevent users from entering incorrect/ill-formatted cuisine types).
  - using enums is also good since the FE can maintain the same list of enums (or we can have an endpoint in BE returning all enums) and we don't have to waste a database transaction to query all the possible enums.

_Things I did not do for this PR_
- set max length of strings in the database
- make sure certain columns cannot be updated (`Promotion.dateAdded`, `Promotion.expirationDate`, etc.)
- upvote/downvote functionality
- restaurant table
- add validation logic
  - e.g. Discount.discountValue cannot have negative value, `expirationDate` has to be strictly greater than `dateAdded`
  - https://typeorm.io/#/validation
  - simple validation https://typeorm.io/#/decorator-reference/check
- did not finish configuring lazy/eager loading
- setup the migrations
- configure all indices

_Cons of TypeORM_
- when I was using `repository.insert([promotion1, promotion2, promotion3])` rather than individually saving each one, the cascades were not being saved correctly. This is a TypeORM bug I believe
  - **solution:** use `repository.save`
- when TypeORM has to create the tables from scratch, and then I call all the `repository.find()` methods, sometimes the repository returns the wrong entity. This is only the case when we have to create the tables from scratch (since `synchronize` is set to true in `ormconfig.json`, see below), but I believe once we arrive at a more concrete decision with our database schema, we won't be making as many drastic database schema changes
  - **solutions:** If we see this is happening too often, we can use the `QueryBuilder` rather than repositories to search for entities. We can also see if this behaviour keeps happening after our DB schema is finalized and we no longer need these two lines
https://github.com/ubclaunchpad/foodies/blob/408e111a8016ea047eefeca02a6849e5b906af6d/backend/ormconfig.json#L8-L9
  - there may be some more bugs related to `repository.find`, but these can be resolved by using the query builder
  - there may be some more bugs with cascade saves, I think I read somewhere this can be solved by using `create` and then `save`, or again, we can just use the raw SQL at the end.
With that said, I can't guarantee TypeORM will be bug-free, but I'm pretty sure these bugs have been discovered by many other people first, and we just have to research. With our use case as well, our schema is very tiny and not too complicated, so I don't think there will be that much trouble. But if TypeORM ends up being more of a trouble than raw SQL, and by then we're already using TypeORM, I can take the time to change our project to raw SQL.

_Things to be wary of:_
- https://typeorm.io/#/relations-faq/avoid-relation-property-initializers
- `ormconfig.json` can actually be a .env file if we want

_Cascade deletes_
- what we have configured right now (certainly not all the cases)
  - if you delete a user, this deletes all their uploaded promotions in the `Promotion` table and then deletes the promotions `discounts`. But if a user has a saved promotion that isn't uploaded by them, then that promotion won't be deleted
  - if you delete a promotion, you delete all the promotions discounts
  - deleting discount doesn't  delete promotion ( we should actually prohibit deletion of discounts unless the promotion is the one that triggered the deletion)
  - deleting promotion doesn't delete user (same reasoning above)
  - deleting promotion also removes any user's who reference the promotion in saved/uploaded.


backend/src/App.ts
```diff
@@ -1,10 +1,75 @@
...
+        // persist entities into database
+        await userRepository.save(user1);
+        await userRepository.save(user2);
+        await promotionRepository.save(promotion1);
+        await promotionRepository.save(promotion2);
+        await promotionRepository.save(promotion3);
+
+        // if user1 wants to add promotion to savedPromotions, this works because of { cascade: true } in User,
+        user1.savedPromotions = [promotion1];
+        await userRepository.save(user1);
```
ericrkuo, line 46: this can also be done using query builder, https://typeorm.io/#/update-query-builder
  - also see definition of `repository.save`: https://typeorm.io/#/entities -  "If id/ids are found then it will update this row in the database. If there is no row with the id/ids, a new row will be inserted."

backend/src/App.ts
```diff
@@ -1,10 +1,75 @@
...
+        const user2 = new User("user2", "test", "user2Email", "user2_userName", "user2_password");
+
+        const promotion1 = new Promotion(user1, [discount3], PromotionCategory.BOGO, CuisineType.CARIBBEAN, "promo1", "test", new Date(), "$")
+        const promotion2 = new Promotion(user2, [discount1, discount2], PromotionCategory.HAPPY_HOUR, CuisineType.JAPANESE, "promo2", "test", new Date(), "$")
+        const promotion3 = new Promotion(user1, [discount4], PromotionCategory.BOGO, CuisineType.KOREAN, "promo3", "test", new Date(), "$")
+
+        // todo: create custom repositories
+        let userRepository: Repository<User> = getRepository(User);
+        let promotionRepository: Repository<Promotion> = getRepository(Promotion);
+        let discountRepository: Repository<Discount> = getRepository(Discount);
```
ericrkuo, line 35: I would recommend https://typeorm.io/#/custom-repository to allow for custom behaviour, e.g. we could implement functions like `customPromotionRepository.getByNameAndCategory`.

backend/src/App.ts
```diff
@@ -1,10 +1,62 @@
...
+        }
+
+        for (let promotion of promotions_sample) {
+            await promotionRepository.save(promotion)
+        }
+
+        for (let [user, promotions] of saved_promotions_mapping) {
+            user.savedPromotions = promotions;
+            await userRepository.save(user);
+        }
```
ericrkuo, line 33: We can either use this to load the data, or run psql command, see the read.me for backend
https://github.com/ubclaunchpad/foodies/blob/backend-orm/backend/README.md#to-import-data-from-init_datasql-run-this-command-inside-a-command-prompt-make-sure-the-schema-is-setup-and-it-is-currently-empty

image

In this PR, there are two ways we can load data, either through 
1. https://github.com/ubclaunchpad/foodies/blob/backend-orm/backend/src/resources/Data.ts
2. https://github.com/ubclaunchpad/foodies/blob/backend-orm/backend/src/resources/init_data.sql
  - credits to @yunishimura0716 for the sample data

backend/README.md
```diff
@@ -1 +1,18 @@
...
+## Foodies Backend Project
+
+Requirements
+1. Postgres
+2. TypeORM Global Installation (`yarn add typeorm -g`)
+    - this will install the TypeORM CLI as well
+
+##### To import data from `init_data.sql`, run this command inside a command prompt. Make sure the schema is setup and it is currently empty
+
+> `psql -d foodies -f resources/init_data.sql --username={YOUR_USERNAME}`
```
andrewzulaybar, line 10:
```suggestion
> `psql -d foodies -f src/resources/init_data.sql --username={YOUR_USERNAME}`
```

Just thought I'd add `src` assuming that people will be running this directly from within `backend/`

backend/src/data/CuisineType.ts
```diff
@@ -0,0 +1,8 @@
+export enum CuisineType {
```
andrewzulaybar, line 1: I agree -- having these as enums are a lot better and will probably be easier to work with. Good choice :)

backend/src/entity/Discount.ts
```diff
@@ -0,0 +1,46 @@
...
+        this.type = type;
+        this.discountValue = discountValue;
+    }
+
+    @PrimaryGeneratedColumn("uuid")
+    id: number;
+
+    // //todo: consider this more (if creating new promotion, would have to search for existing discounts?)
+    // @ManyToMany(() => Promotion, promotion => promotion.discounts)
+    // promotions: Promotion[];
```
andrewzulaybar, line 22: Yeah, we can remove this. I like your suggestion from the Google doc

backend/src/entity/Promotion.ts
```diff
@@ -0,0 +1,104 @@
...
+    * Promotion is the owning side of this association, contains column userId
+    * */
+    @ManyToOne(() => User, user => user.uploadedPromotions, {
+        nullable: false,
+        onDelete: "CASCADE"
+    })
+    @JoinColumn()
+    user: User;
+
+    // @ManyToMany(()=> Discount, discount => discount.promotions, {
```
andrewzulaybar, line 48: We can remove this too now!

andrewzulaybar (approved):
This is amazing @ericrkuo!!! 🙌

I'm really looking forward to using this. I think TypeORM was a good choice and it looks like it has a bunch of promising features that I hope we can make use of. Thanks so much for all your hard work on this 🌟

yunishimura0716 (approved):
It's awesome work!. TypeORM make us easy to control db.

backend/README.md
ericrkuo: Sounds good!

backend/src/data/CuisineType.ts
ericrkuo: YUp and one nice thing is that postgres has enum types for columns, so if anyone tries to insert any `cuisinetype`/`discountType`/`promotionCategory` that isn't part of the enum, postgres will complain.

backend/package.json
```diff
@@ -10,6 +10,10 @@
     "test": "jest",
     "lint": "eslint . --ext .ts,.tsx --fix",
     "prettier": "prettier --write ."
+    "test": "jest",
```
ld357, line 13: Is this the same as line 10?

backend/src/entity/Discount.ts
```diff
@@ -0,0 +1,42 @@
+import {Entity, PrimaryGeneratedColumn, Column, ManyToOne, JoinColumn} from "typeorm";
+import {Promotion} from "./Promotion";
+import {DiscountType} from "../data/DiscountType";
+
+/*
+* Represents a discount to a promotion. Each discount has a DiscountType which determines the discountValue.
```
ld357, line 6: I can't remember if we talked about this 😅  but for documenting functions etc, do we want to do this for everything and go with a specific library (tsdoc, etc)?  @andrewzulaybar @ericrkuo

ld357 (approved):
Awesome work! 🎉

backend/package.json
ericrkuo: fooooooook i messed up the rebase

ericrkuo: Thankss ill fix that

backend/src/entity/Discount.ts
ericrkuo: yeyeye I would prefer documentation library. That way we can use annotations like `@param`, `@return`. And when you hover over it it looks nice and fancy. I can add it in this PR if its okay with @andrewzulaybar?

backend/src/entity/SavedPromotion.ts
```diff
@@ -0,0 +1,41 @@
...
+import {User} from "./User";
+import {Promotion} from "./Promotion";
+
+/*
+* Represents a user who saved a promotion
+* * the join/association table which connects User and Promotion
+* * contains custom properties
+* */
+@Entity()
+export class SavedPromotion {
```
ericrkuo, line 11: This is a join/association table that has custom properties. Now whenever we add a user adds a saved promotion, the column `dateSaved` automatically generates the timestamp of when the user saved.

See https://typeorm.io/#/many-to-many-relations/many-to-many-relations-with-custom-properties

backend/src/resources/init_data.sql
```diff
@@ -0,0 +1,113 @@
+INSERT INTO public."user" (id, first_name, last_name, email, username, password)
```
ericrkuo, line 1: I can make a GIF or add to the `readMe` how to generate these from Intellij

backend/src/entity/Discount.ts
andrewzulaybar: @ld357 Yeah good suggestion! I'd like to have one too, probably tsdoc. @ericrkuo I just remembered you brought up https://typedoc.org/ and I forgot to comment on it -- I think we can use this if it's easy enough to work with, but I also think our project isn't big enough yet that we should be able to add these comments in manually without the need of adding another library? But maybe it might help with generating docs for the future? But what're your thoughts? Also, could you add the documentation in another PR either branching off of this one or off of master once you merge this in? It'll just be easier to review that way :)

ericrkuo: Oh yeah I think we can stick with manual comments like this for now! Doesn't really matter that much to me, and yup sure I can make a separate PR for it if we decide to use it

richardyjtian:
Great work @ericrkuo! Yes I think TypeORM makes everything a lot easier to read compared to raw SQL. Wish I'd discovered this earlier too.

richardyjtian (approved):

backend/test/service/GooglePlacesService.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  beforeEach(function () {
+    dotenv.config();
+  });
+
+  test('Search for a place using restaurant name, should be successful', () => {
+    return googlePlacesAPI
+      .getGooglePlacesSearch('Jinya')
+      .then((result: JSON) => {
+        // todo: need some assertions
```
ericrkuo, line 18: I had to change this from `AxiosResponse<JSON>` to `JSON` because the compiler was giving me an error. Also had to change `dotenv` to an import rather than a require.

I made the `test` folder structure the same as the `src` folder structure, so the unit test for `GooglePlacesService` would be in the same `service` folder inside `test`. It's something I find kind of useful, I can change it back if other people don't want it to be like that.

backend/sample.env
```diff
@@ -0,0 +1 @@
+GOOGLE_PLACES_API_KEY=akjshdaksjhdjk
```
ericrkuo, line 1: This way we can monitor all our env variables without having to worry about accidentally commiting our `.env` file and everyone can be up to date with any environment variables they are missing. ReadME has some more instructions!

backend/README.md
```diff
@@ -17,6 +17,23 @@ Make a copy of `sample.env` and rename it to `.env`.
 Fill out the environment variables respectively
 ```
 
+## Setting up databases
+
+You will need two databases, `foodies` and `foodies_test`
```
ericrkuo, line 22: Created a database for writing unit/integration tests.

backend/ormconfig.json
```diff
@@ -8,12 +8,12 @@
   "synchronize": true,
   "dropSchema": true,
   "logging": false,
-  "entities": ["src/entity/**/*.ts"],
-  "migrations": ["src/migration/**/*.ts"],
-  "subscribers": ["src/subscriber/**/*.ts"],
+  "entities": ["src/main/entity/**/*.ts"],
```
ericrkuo, line 11: Made folder structure to be like this:
image

backend/src/main/repository/UserRepository.ts
```diff
@@ -0,0 +1,40 @@
+import { EntityRepository, Repository } from 'typeorm';
+import { User } from '../entity/User';
+import { Promotion } from '../entity/Promotion';
+import { SavedPromotion } from '../entity/SavedPromotion';
+
+@EntityRepository(User)
+export class UserRepository extends Repository<User> {
```
ericrkuo, line 7: Now we are using custom repositories, they can implement functions like below!

backend/src/test/repository/UserRepository.test.ts
```diff
@@ -0,0 +1,25 @@
+import { getConnection, getCustomRepository } from 'typeorm';
+import { User } from '../../main/entity/User';
+import { users_sample } from '../../main/resources/Data';
+import { UserRepository } from '../../main/repository/UserRepository';
+import { BaseRepositoryTest } from './BaseRepositoryTest';
+
+describe('Unit tests for UserRepository', function () {
+  beforeEach(() => {
+    return BaseRepositoryTest.establishTestConnection();
```
ericrkuo, line 9: This will establish a connection with our `foodies_test` database each time, so it won't interfere with `foodies` database. 

Originally, I wanted to create a test hierarchy with the base class being `BaseRepositoryTest`. But this was a quicker solution with a static method.

Issue #17, opened 2020-10-25
Create card components for promotions


andrewzulaybar:
We want to be able to list and view either restaurants or promotions (to be discussed with design), so we need to implement the card components that will hold this information.

We'll need to discuss with design exactly what information we want to display in these cards.

andrewzulaybar:
@wendoz @ElinaJiang Tagging you both as a reminder to myself that I need to talk about this more!

Issue #33, opened 2020-11-01
Add search bar for promotions


andrewzulaybar:
## Task 🔨
Add search bar for top navigation bar that allows users to search for a specific promotion.

<img width="352" alt="Screen Shot 2020-10-31 at 5 47 34 PM" src="https://user-images.githubusercontent.com/44531733/97792859-53876980-1ba1-11eb-9139-efa0041b56ce.png">

Issue #32, opened 2020-11-01
Add buttons for sorting and filtering list of promotions to Home page 


andrewzulaybar:
## Task 🔨
Add the 6 buttons in the photo below to manage sorting and filtering of the list of promotions. Exact categories for the dropdowns are still TBD.

<img width="722" alt="Screen Shot 2020-10-31 at 5 44 32 PM" src="https://user-images.githubusercontent.com/44531733/97792833-e673d400-1ba0-11eb-8801-0fc49c1f890e.png">

PR #47, opened 2020-11-06
Add skeleton implementation of Home Page

andrewzulaybar:
Fixes: #17, #32, #33

Added a bare-bones implementation of the front-end according to the current design. This PR includes the following features:
- Navigation bar with `react-router-dom` to allow navigation between routes
- Search bar, currently just alerts the user if an input is entered
- Dropdown buttons for filtering the list of promotions, each with a list of options (concrete options still TBD)
- Promotion detail cards

The Map component that @richardyjtian created in #22 still needs to be integrated, will do this in another PR.


ld357 (approved):
Thanks for letting me review also! 😄

engineereng (approved):
Great work!

richardyjtian (approved):
Looks good!

Issue #48, opened 2020-11-07
Add validation to our backend


ericrkuo:
We need to add validation logic to our database to prevent faulty data, but before we do that we need to research how we are going to approach this.

TypeORM has certain validation features we can leverage
- https://typeorm.io/#/validation
- https://typeorm.io/#/decorator-reference/check
- https://github.com/typeorm/typeorm/issues/913#issuecomment-533853309
- Maybe we can use something with subscribers/listeners

We should also consider doing some validation that is already covered by the database. For example, the database currently has a unique column for `User.username`, but rather than inserting a new user with the same username and letting Postgres generate that error, we should check it beforehand so we don't waste a database transaction and we can have our own custom error messages.

This ticket would then also be responsible for making sure all our errors returned from our controllers follow a similar format.
For example, we could have something like this:
```
FrontEndErrorObject{
  errorCode: 'UsernameAlreadyTaken',
  message: ["mesg1", "msg2"],
  debug: "how to debug"
}
```

Then we could have an `Error` class representing all our global errors.

Some other validations to consider:
- set max length of strings in the database
- make sure certain columns cannot be updated (`Promotion.dateAdded`, `Promotion.expirationDate`, etc.)
- `Discount.discountValue` cannot have negative values
- `expirationDate` has to be strictly greater than `dateAdded` for a promotion
- each time a user uploads a promotion, check the validity of the promotion
- each time a user is created, check the validity of a user
- each time discount is added, check the validity of discount

Issue #40, opened 2020-11-01
Add controller for DELETE /promotion/:promotion_id


andrewzulaybar:
## Task 🔨

The request body could include:
- user_id (refer to note in #37)

This controller should be able to:
- delete entry in Promotion table
- delete the entries in the related tables: Promotion_Has_Category, Category, Promotion_Has_Discount, Discount but this should all be handled by specifying the `ON DELETE` property of the SQL table

The route should respond with:
- status code 204 if successfully deleted (should be idempotent so the deletion of resource should respond the same whether it exists in the database or not)

Issue #39, opened 2020-11-01
Add controller for POST /promotion/:promotion_id


andrewzulaybar:
## Task 🔨

This controller should be able to:
- retrieve the promotion from the Promotion table (joining any other tables if necessary) given the `promotion_id` parameter in the URI

The route should respond with:
- status code 200 if successfully found
- status code 404 if not found

ericrkuo:
I'll be working on this issue along with https://github.com/ubclaunchpad/foodies/issues/48 at the same time since they're pretty connected.

Issue #38, opened 2020-11-01
Add controller for POST /promotion


andrewzulaybar:
## Task 🔨

The request body could include:
- user_id (refer to note in #37)
- all fields for a valid promotion, according to schema

This controller should be able to:
- add a new entry to the `Promotion` table
- add a new entry to the `Promotion_Has_Category` table, and optionally the `Category` table if it doesn't exist
- add a new entry to the `Promotion_Has_Discount` table, and optionally the `Discount` table if it doesn't exist

The route should respond with:
- status code `201` or `204` if successfully added
- status code `422` if we're unable to process the request

ericrkuo:
I'll be working on this issue along with https://github.com/ubclaunchpad/foodies/issues/48 at the same time since they're pretty connected.

PR #52, opened 2020-11-12
Added endpoints for promotions, validation, error handling

ericrkuo:
**Issues:**
https://github.com/ubclaunchpad/foodies/issues/36 https://github.com/ubclaunchpad/foodies/issues/37 https://github.com/ubclaunchpad/foodies/issues/38 https://github.com/ubclaunchpad/foodies/issues/39 https://github.com/ubclaunchpad/foodies/issues/40 https://github.com/ubclaunchpad/foodies/issues/48

**New folders:**
image
- constant - contains any constants we use in project
- controller - implements controllers for routes
- middleware - any middleware we will use for express (for example see https://github.com/ubclaunchpad/foodies/blob/backend-promotion/backend/src/main/middleware/ErrorHandler.ts)
- route - configures the routes
- validation - any validation logic we will be adding

**Summary:**
There are fourtmain components in this PR, the Promotions endpoint, validation, error handling, and Enum endpoints.

**Validation:**
We needed a way to validate any data we get from a client (for example the request body for POST request to `promotions`, or query params for GET request to `promotions`). We cannot just cast it to an interface because the types may be completely wrong.

The solution I did was to use Joi (https://joi.dev/api/?v=17.3.0#introduction). This library has a very impressive number of downloads (2 million). It is very simple to set up, all we need to define is a schema to validate against.

For example, say we want to validate the contents of the request body for a POST request. We can't just cast it to an interface because the contents of it could still be invalid, see https://stackoverflow.com/questions/43894565/cast-object-to-interface-in-typescript or https://stackoverflow.com/questions/55937458/typescript-convert-post-request-body-to-map#comment104517695_55937659. Instead, we can do something like this:

https://github.com/ubclaunchpad/foodies/blob/6eb159a5c29916d0b99331ad452432d13b1c7c03/backend/src/main/controller/PromotionController.ts#L135-L138.

First, we call `PromotionValidation.schema` to access the `Joi` schema (static property) and to validate it against the `request.body`. Then we cast it to `promotionDTO` since we know now that if validation succeeds, we have the correct types and format. If it's confusing, message me for clarification or the Joi docs also explain validation pretty well, but the main point is that we validate and cast). If you want to see the validation schema, checkout https://github.com/ubclaunchpad/foodies/blob/6eb159a5c29916d0b99331ad452432d13b1c7c03/backend/src/main/validation/PromotionValidation.ts#L11-L25

**Promotion endpoints**
Here, we register all our routers, https://github.com/ubclaunchpad/foodies/blob/6eb159a5c29916d0b99331ad452432d13b1c7c03/backend/src/main/App.ts#L43-L49. The reason why I instantiate things here is for dependency injection, we can use a DI library with a container to clean things up in the future.

Then we configure all our routes and assign them to the necessary controller methods
https://github.com/ubclaunchpad/foodies/blob/6eb159a5c29916d0b99331ad452432d13b1c7c03/backend/src/main/route/PromotionRouter.ts#L11-L20

**One important note:** I made it so that controllers communicate with the databases. If we need any business logic, we would put those into services and call them from controllers. Let me know your thoughts on this.

**Error Handling**
There is a new middleware handler for any errors:
https://github.com/ubclaunchpad/foodies/blob/6eb159a5c29916d0b99331ad452432d13b1c7c03/backend/src/main/App.ts#L33

The handler is global, any error thrown is handled by this class. As you can see, we can also customize the status code and messages depending on the type of error thrown.
https://github.com/ubclaunchpad/foodies/blob/6eb159a5c29916d0b99331ad452432d13b1c7c03/backend/src/main/middleware/ErrorHandler.ts#L18-L31

Each error is then converted into a `FrontEndErrorObject`. This basically ensures consistency in how we communicate back to the FE.
https://github.com/ubclaunchpad/foodies/blob/6eb159a5c29916d0b99331ad452432d13b1c7c03/backend/src/main/data/FrontEndErrorObject.ts#L1-L6

**Enum endpoints**
Added enum endpoints, see https://github.com/ubclaunchpad/foodies/pull/52#discussion_r523380537

**What I didn't do**
- Create tests for the controllers
- for GET promotions, I didn't add sorting. I thought it would make more sense for FE to sort it on their end. @andrewzulaybar is that okay?


backend/ormconfig.json
```diff
@@ -6,7 +6,7 @@
   "password": "postgres",
   "database": "foodies",
   "synchronize": true,
-  "dropSchema": true,
+  "dropSchema": false,
```
ericrkuo, line 9: Decided that its best we don't drop the schema whenever we start the application. If local devs want it to then they can set it back to true!

backend/src/main/App.ts
```diff
@@ -23,62 +27,11 @@ createConnection()
...
-    const promotionRepository: PromotionRepository = getCustomRepository(
-      PromotionRepository
-    );
-    const discountRepository: DiscountRepository = getCustomRepository(
-      DiscountRepository
-    );
-    const savedPromotionRepository: SavedPromotionRepository = getCustomRepository(
-      SavedPromotionRepository
-    );
+    // await loadSampleData();
```
ericrkuo, line 30: Commented this out for now, for me when developing locally it was easier to load the data only **once.** So synchornize is true, dropSchema is false. That way we can leverage nodemon, otherwise nodemon would keep executing this line everytime we save.

backend/src/main/controller/PromotionController.ts
```diff
@@ -0,0 +1,188 @@
...
+          request.query,
+          { abortEarly: false }
+        );
+        queryBuilder = this.applyQueryOptions(queryBuilder, promotionQuery);
+      }
+
+      const promotions = await queryBuilder.getMany();
+      return response.send(promotions);
+    } catch (e) {
+      return next(e);
```
ericrkuo, line 46: this will then go to the error handler middleware

backend/src/main/controller/PromotionController.ts
```diff
@@ -0,0 +1,188 @@
...
+   */
+  getAllPromotions = async (
+    request: Request,
+    response: Response,
+    next: NextFunction
+  ): Promise<any> => {
+    try {
+      let queryBuilder = getCustomRepository(PromotionRepository)
+        .createQueryBuilder('promotion')
+        .innerJoinAndSelect('promotion.discount', 'discount');
```
ericrkuo, line 33: For this endpoint I could not use a repository, unfortunately. But using the queryBuilder isn't that complicated. The reaosn why we need query builder is that if we want to filter by discount type, then we need to join. 

We cannot just do 
```
getCustomRepository(PromotionRepository).find({where: discount {type: '%'}, relations: ['discount']}
````

File: backend/src/main/controller/PromotionController.ts [hunk 5]
```diff
@@ -0,0 +1,188 @@
...
+      return response.send(promotions);
+    } catch (e) {
+      return next(e);
+    }
+  };
+
+  /**
+   * Depending on which properties are defined inside promotionQuery, we add those properties into our query for the queryBuilder to execute.
+   */
+  private applyQueryOptions(
```
ericrkuo, line 53: This function attaches more to the query builder depending on any filters we apply for `GET promotions`.

backend/src/main/entity/User.ts
```diff
@@ -67,6 +67,7 @@ export class User {
   // todo: we need something more secure
   @Column({
     unique: true,
+    select: false, // https://typeorm.io/#/select-query-builder/hidden-columns
```
ericrkuo, line 70: Anytime we access a user, we don't want to get their password. If in any case we need a user's password, we need to use a query builder to specifically ask for the password

backend/src/main/controller/PromotionController.ts
```diff
@@ -0,0 +1,188 @@
...
+      // this way users can query promotions that contain % or _
+      name = name.replace('%', '\\%');
+      name = name.replace('_', '\\_');
+
+      // add wildcards
+      name = `%${name.replace(/ /g, '%')}%`;
+
+      queryBuilder = queryBuilder.andWhere('promotion.name ilike :name', {
+        name: name,
+      });
```
ericrkuo, line 97: So this is how I'm thinking we could the FE submits queries for promotion names.

Say for example the user wants to find promotions that contain "promo hi". We would then add wildcards `%` at the start and end of the string, and in spaces. That way we can leverage `postgres`'s `ILIKE` comparator.

In cases where the user is searching for `50%`, then we don't want the % to act as a wildcard. THis will be the responsibility of the FE to encode % into `%25` (see the link in the code). That way we can escape % by turning it into `\\%`, then adding the wildcards.

@na-wu would there be a better way for this search query?

backend/src/main/route/PromotionRouter.ts
```diff
@@ -0,0 +1,21 @@
...
+import { PromotionController } from '../controller/PromotionController';
+
+export class PromotionRouter {
+  private promotionRouter = express.Router();
+  private promotionController = new PromotionController();
+  constructor(promotionController: PromotionController) {
+    this.promotionController = promotionController;
+  }
+
+  getRoutes(): Router {
```
ericrkuo, line 11: @yunishimura0716 let me know if there's anything in this PR you'd want to follow a different design, for example the routers, controllers, services, validation etc. I know you are working on the `User`'s endpoints at the moment and it will take some coordination from both of us to make sure our PRs follow the same format and don't have merge conflicts. 

@andrewzulaybar 
I did not implement the following endpoints from the meeting notes. I feel like these are endpoints meant for the `UserController`. We also need an endpoint to `GET` the saved promotions.
```
POST /promotion/:promotion_id/save
DELETE /promotion/:promotion_id/save
```

I was thinking more along the lines of:
```
POST /users/:id/savedPromotions
DELETE /users/:id/savedPromotions
GET /users/:id/savedPromotions
```

ericrkuo (commented):
Open to feedback in anything I can improve! Let me know if there's anything you want to change 😃

backend/src/main/controller/PromotionController.ts
```diff
@@ -0,0 +1,188 @@
...
+  ): Promise<any> => {
+    try {
+      let queryBuilder = getCustomRepository(PromotionRepository)
+        .createQueryBuilder('promotion')
+        .innerJoinAndSelect('promotion.discount', 'discount');
+
+      if (request.query) {
+        const promotionQuery: PromotionQueryDTO = await PromotionQueryValidation.schema.validateAsync(
+          request.query,
+          { abortEarly: false }
```
ericrkuo, line 38: `abortEarly: false` basically tells Joi to continue validating even if there is already an error. This way it'll give us back ALL the validation issues at once, so if multiple parts of the request are ill-formatted itll tell us all parts at once.

For example, if the incoming cuisine type isn't a valid `CuisiineType` and the expiration_date isn't a valid date, it'll give us back both validation errors rather than only the first one it encounters

backend/src/main/controller/PromotionController.ts
ericrkuo: could be moved to `PromotionService` class, but I thought since its relevant to communicating w DB we can leave it here

yunishimura0716 (approved):
Great work Eric! And it's very scalable design!

backend/src/main/controller/EnumController.ts
```diff
@@ -0,0 +1,64 @@
...
+        request.params.enum,
+        { abortEarly: false }
+      );
+      const enumClass = this.getEnumClass(enumString);
+      const enumValues = this.getEnumValues(enumClass);
+      return response.status(200).send(enumValues);
+    } catch (e) {
+      next(e);
+    }
+  };
```
ericrkuo, line 39: @andrewzulaybar I remember you said you wanted enums to ideally be exposed through endpoints.

This controller now allows us to call `enums/:enum` and will give us all the possible values of the specified enum.

So we have
```
GET enum/DiscountType
GET enum/CuisineType
GET enum/PromotionCategory
```

backend/src/main/controller/EnumController.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * returns the corresponding class matching the enumString
+   * */
+  private getEnumClass(supportedEnums: SupportedEnums) {
+    switch (supportedEnums) {
```
ericrkuo, line 45: Unfortunately had to use switch statement. Does anyone have any suggestions? Typescript won't let me map to type `enum`.

I basically want to map a string to an `enum`

backend/src/main/controller/PromotionController.ts
```diff
@@ -0,0 +1,121 @@
...
+   */
+  getAllPromotions = async (
+    request: Request,
+    response: Response,
+    next: NextFunction
+  ): Promise<any> => {
+    try {
+      const promotions = await getCustomRepository(
+        PromotionRepository
+      ).getAllPromotions(request.query);
```
ericrkuo, line 29: @andrewzulaybar 
This is kind of what the search query is like atm. You can add a number of request query params and they're all an AND clause. I plan on working a bit more on searching by `name`, but this is what I have so far.

image

backend/ormconfig.json
```diff
@@ -10,6 +10,7 @@
   "logging": false,
   "entities": ["src/main/entity/**/*.ts"],
   "migrations": ["src/main/migration/**/*.ts"],
+  "migrationsRun": true,
```
ericrkuo, line 13: Basically, this will run any migrations inside `src/main/migration`.
https://typeorm.io/#/migrations

backend/src/main/entity/Discount.ts
```diff
@@ -21,7 +21,7 @@ export class Discount {
   }
 
   @PrimaryGeneratedColumn('uuid')
-  id: number;
+  id: string;
```
ericrkuo, line 24: Made a mistake, all `id` columns should be strings.

backend/src/main/entity/Promotion.ts
```diff
@@ -131,4 +131,23 @@ export class Promotion {
...
+    nullable: false,
+    select: false,
+  })
+  tsVector: string;
+
+  /**
+   * Used only for full text search when returning rank back to client.
+   * Represents how relevant documents are to a particular query, so that the most relevant one can be shown
+   */
+  rank?: number;
```
ericrkuo, line 152: This part will be difficult to understand unless you study https://www.postgresql.org/docs/13/textsearch.html. But basically a `tsvector` in Postgres is used for full text search.

backend/src/main/repository/PromotionRepository.ts
```diff
@@ -16,66 +11,111 @@ export class PromotionRepository extends Repository<Promotion> {
...
+      const promotion = mapIdToPromotion.get(idRank.id);
+      if (promotion) {
+        promotion.rank = idRank.rank;
+        result.push(promotion);
+      }
+    }
+
+    return result;
   }
 }
```
ericrkuo: This is still a rough algorithm, but here is what's happening
1. `promotionQuery.searchQuery` will contain the keywords we want to search up. (e.g. "Vancouver Chinese dumplings")
2. `arrayOfIdRank` will break down the search query into a tsvector consisted of lexemes and the search query will be using `OR` operators.
     - it will then return a promotionId along with the rank, where the rank denotes how well the promotion name/description maps to the search query
3. `promotions` will contain the result of finishing the rest of the query builder (which may contain `where` clauses for discount/category/expirationdate/etc. and joining with discount, but it will only take promotions that have id inside `arrayOfIdRank`.
4. We add `rank` value to each promotion and return.

Let me know if any of this needs further clarification, I'd be more than happy to discuss!

ericrkuo (commented):
https://github.com/ubclaunchpad/foodies/pull/52/commits/0f1728b9c40010d69457f95babf05b63100f1524
Take a look at this commit to learn about full-text search in Postgres. This will basically enhance our search query.

I'd recommend reading this, https://www.compose.com/articles/mastering-postgresql-tools-full-text-search-and-phrase-search/, otherwise it will be difficult to understand.

If you want to learn it more in detail, here is a better link! https://www.postgresql.org/docs/13/textsearch.html

I've optimized the full text search by adding in a column inside `promotion` table called `tsvector` to store the tsvector inside. This required creating a GIN index and also triggers to listen for inserts and updates.

backend/src/main/controller/PromotionController.ts
ericrkuo: Edit: disregard this, I tried to implement using postgres's full text search. @na-wu lmk your thoughts if you get the chance to take a look! It's mainly this commit https://github.com/ubclaunchpad/foodies/commit/0f1728b9c40010d69457f95babf05b63100f1524

backend/ormconfig.json
ld357: Just curious but what does this mean? 😄

backend/src/main/controller/EnumController.ts
```diff
@@ -0,0 +1,64 @@
...
+import { NextFunction, Request, Response } from 'express';
+import Joi from 'joi';
+
+enum SupportedEnums {
+  DiscountType = 'DiscountType',
+  PromotionCategory = 'PromotionCategory',
+  CuisineType = 'CuisineType',
+}
+
+/* eslint-disable  @typescript-eslint/no-explicit-any */
```
andrewzulaybar, line 13: Feel free to remove this linting rule if it's been more annoying than helpful 😛

backend/src/main/controller/EnumController.ts
andrewzulaybar: This is great, thank you 🙌

backend/src/main/controller/EnumController.ts
andrewzulaybar: I can't think of anything that would be a much better alternative :( You could drop the enclosing brackets for each case to make it a little less verbose though

backend/src/main/controller/PromotionController.ts
```diff
@@ -0,0 +1,131 @@
...
+      });
+      const promotion = await getCustomRepository(PromotionRepository).delete(
+        id
+      );
+      return response.status(204).send(promotion);
+    } catch (e) {
+      return next(e);
+    }
+  };
+}
```
andrewzulaybar, line 131: Love how clean and easy to read these are, great job @ericrkuo

backend/src/main/repository/PromotionRepository.ts
```diff
@@ -1,5 +1,121 @@
...
+      return this.createQueryBuilder('promotion')
+        .innerJoinAndSelect('promotion.discount', 'discount')
+        .getMany();
+    }
+  }
+
+  /**
+   * Depending on which properties are defined inside promotionQuery, we add those properties into our query for the queryBuilder to execute.
+   */
+  private applyQueryOptions(promotionQuery: PromotionQueryDTO): Promise<any> {
```
andrewzulaybar, line 30: I thought about this a bit more and was wondering if you also agree the filtering/sorting should be done server-side? I thought that it would be nice to do the queries as close to the DB as possible, but I forgot that that means making a network request each time the user wants to filter/sort. Another thought that I had is that we'd be displaying the promotions within each filter ordered by distance (my assumption), so that would mean we'd have to do processing work on the FE anyways to retrieve lat/lon data. But maybe you have other thoughts on this?

backend/src/main/route/PromotionRouter.ts
andrewzulaybar: @ericrkuo That sounds good! So like the request body to `POST` and `DELETE` `/users/:id/savedPromotions` would include the promotion_id? Also another thing that I'm not sure I considered earlier, have we designed a way to get the number of saves for a specific promotion?

andrewzulaybar (approved):
This looks amazing! Thanks for the hard work on this 🙌

I just left a few questions about some discussion points. Feel free to merge this in when you're ready!

backend/src/main/controller/EnumController.ts
ericrkuo: hehe okay thank you 😅

backend/src/main/controller/EnumController.ts
ericrkuo: Ohya for sure

backend/src/main/repository/PromotionRepository.ts
ericrkuo: yeahh like this is def something I wanted to bring up, I think I like sorting in the front end, that way it can be much faster without the need to make a request each time. (like if we perform a complex search query + filters, and get the results, then the user wants to sort, we will need to execute the complex search query + filters again in addition to the sort)

But I think filtering itself can be a network request, mainly because of what you mentioned about queries as close to the DB as possible. https://stackoverflow.com/questions/52346685/filters-logic-should-be-on-frontend-or-backend#:~:text=This%20depends%20on%20the%20size,end%20after%20getting%20the%20data.

At the same time, I've attached the `rank` from PostgreSQL's full-text search, so you still have the info on which promotions are most relevant to the search query so that the FE can sort by relevance (rank).

ericrkuo: Let me know what you think about this!

backend/src/main/route/PromotionRouter.ts
ericrkuo: yup! so promotionId would be a request query, and id would be request param.

Yu is working on it currently, but I remember when I was playing around with it, you could do this:
image

@yunishimura0716 the very tricky part is getting the `savedAt` timestamp from the `saved_promotion` table. I think I have an idea of how to do this.
1. use saved promotion repository to get all entries where userId is the desired id. (do not do any joins, we just need the promotionId's and the savedAt timestamp)
2. use all the promotionIds from 1, and execute query to get all promotions + join with discount where id is in the id's from 1.
3. then you can probably do something like
```
return promotions.map((promotion) => return {
  ...promotion,
  savedAt: <the respective timestamp>
}
```
 The reason why you cannot just execute query is because the data won't be formatted the way we want to
we want:
```
we want:
{
  id: 
  name: 
  discount: {
     type:
  }
  savedAt:
}

but we'll actaully get 

{
  id: 
  name: 
  discount_type:
  discount_discountValue:
  savedAt
}
```

This is because if you want to execute any sql query that doesn't return a promotion (because the promotion entity doesn't have a column `savedAt`), you need to call `getRawMany`, and it won't be formatted the way we want it to. This might not make sense, but I think you'll encounter this problem when you try to get the `savedAt` time stamp.

Here's how I did it for full text search, but I added a private property to `Promotion` called `rank` instead of doing the `.map(promotion => return {...promotion, savedAT: ...})` approach.

ld357 (approved):
Amazing job Eric!!! 👍🏻  💯

backend/src/main/repository/PromotionRepository.ts
ericrkuo: Ignore: will explain this in other issue that concerns fully implementing search query

backend/src/main/repository/PromotionRepository.ts
andrewzulaybar: @ericrkuo Yeah that makes sense! Also, filtering on BE is nice because it minimizes the packet sizes of our network requests (although this probably won't be a big problem for us since we're not dealing with big data). I think we'll have to do some combination of sorting by several keys (e.g. rank, distance) depending on different situations but that can all be handled by FE so this works

ericrkuo: Awesome sweet!

ericrkuo: Yah that makes sense about packet sizes. I'm thinking that in the future we will encounter this problem about packet sizes as we will have quite a lot of data to send back. But I think it should be okay hopefully for now.

PR #56, opened 2020-11-14
Added endpoints for user profile

yunishimura0716:
I created endpoint for user profile data.
Based on the advise of @ericrkuo, My implementation for route, controller and validation are similar to his way.

**Validation** => using _Joi_

**What I have done not yet**
- createing test for validation (actually I created but I think It's not enougth)
- creating routes and controller for updated and saved promotion of each user

Thanks @ericrkuo for helping, welcome to any comment!!


backend/src/main/entity/User.ts
```diff
@@ -69,4 +70,12 @@ export class User {
     unique: true,
   })
   password: string;
+
+  hashPassword() {
+    this.password = bcrypt.hashSync(this.password, 8);
+  }
```
ericrkuo, line 76: What do you think about something like this?
```
  constructor(
    firstName: string,
    lastName: string,
    email: string,
    username: string,
    password: string
  ) {
    this.firstName = firstName;
    this.lastName = lastName;
    this.email = email;
    this.username = username;
    this.password =bycrypt,hashSync(password, 8); // 
  }
```
This way we don't need to maintain the overhead of encrypting the password each time we create a new user, we can just assume once we create a user the password is encrypted.

ericrkuo: Then we also don't need the changes inside `Data.ts`

backend/src/main/controller/UserController.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    // return one user
+    getOneById = async (req: Request, res: Response, next: NextFunction) => {
+        try {
+            // get the id from the url
+            const id = await IdValidation.schema.validateAsync(req.params.id, {
+                abortEarly: false,
+            });
+            const userRepository = getCustomRepository(UserRepository);
+            const user = await userRepository.findOneOrFail(id, {
```
ericrkuo, line 40: Oh wait for this one, do you want to try to add `select: false`?
https://github.com/ubclaunchpad/foodies/blob/6eb159a5c29916d0b99331ad452432d13b1c7c03/backend/src/main/entity/User.ts#L68-L72

Then you can just do `userRepository.findOneOrFail(id)` without the select.

backend/src/main/controller/UserController.ts
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    // create new user
+    newUser = async (req: Request, res: Response, next: NextFunction) => {
+        try {
+            // get parameters from the body
+            const userDTO: UserDTO = await UserValidation.schema.validateAsync(
+                req.body,
+                { abortEarly: false }
+            );
```
ericrkuo, line 56: Awesome work!!!!

backend/src/main/validation/UserValidation.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * check the validity of an user, used when we make a POST request to /users
+ */
+export class UserValidation {
+    static schema: ObjectSchema = Joi.object({
+        username: Joi.string().required(),
+        firstName: Joi.string().required(),
+        lastName: Joi.string().required(),
+        email: Joi.string().required(),
```
ericrkuo, line 11: I saw that `Joi` has this additional module for email validation: https://joi.dev/module/address/#install.

Do you want to try this out?

I think there also might be `Joi.string().email()`.

ericrkuo (approved):
Very nice work Yu!

backend/src/main/entity/User.ts
yunishimura0716: Yeah, that's good idea! I want to ask you that if we use 'update' method of repository, is this constructor applied?

backend/src/main/validation/UserValidation.ts
yunishimura0716: Ok, I'll try!

backend/src/main/entity/User.ts
ericrkuo: Ah wait that's a good point, then we would have to manually do something like inside `update` method of repository: 
```
if (userUpdateDto.password) {
 userUpdateDto.password = bycript.hashSync(userUpdateDto.password, 8)
}
```

backend/src/main/route/UserRouter.ts
```diff
@@ -0,0 +1,25 @@
+import { Router } from "express";
+import { UserController } from "../controller/UserController";
+
+export class UserRouter {
+    private userRouter = Router();
+    private userController = new UserController();
```
ericrkuo, line 6: Sorry, this was a mistake on my end, shouldn't instantiate with new since will be passed in the constructor.
Can just leave it as `private userController;`. I will fix on my PR

backend/src/main/entity/User.ts
yunishimura0716: Ok, I'll do that!

backend/src/main/App.ts
```diff
@@ -23,6 +26,16 @@ createConnection()
     app.use(bodyParser.json());
     const PORT = 8000;
 
+    registerRouters(app);
+
+    function registerRouters(app: express.Express) {
+      app.get('/', (req, res) => res.send('Hello World'));
```
andrewzulaybar, line 32: We can get rid of this line now I think!

andrewzulaybar (approved):
I added some linting and formatting on master, so you could merge master into this branch and run `yarn lint` and then `yarn prettier` to fix the indentation but otherwise, this looks really good!

yunishimura0716 (commented):
I created the endpoint for savedPromotion. The methods contain GET/POST/DELETE.
Through the discussion with @ericrkuo, we decided to use SavedPromotion Repository for the method, addSavedPromotion.

backend/src/main/controller/UserController.ts
```diff
@@ -0,0 +1,185 @@
...
+            const uid = await IdValidation.schema.validateAsync(req.params.id, {
+                abortEarly: false,
+            });
+            const pid = await IdValidation.schema.validateAsync(req.params.pid, {
+                abortEarly: false,
+            });
+            // create new saved promotion for the user based on promotion id
+            const promotion = await getCustomRepository(PromotionRepository).findOneOrFail(pid);
+            const user = await getCustomRepository(UserRepository).findOneOrFail(uid);
+            const result = await getCustomRepository(UserRepository).addSavedPromotion(user, promotion);
```
ericrkuo, line 138: Hmm, do you know if there is any difference in doing it this way? I'm not sure if its faster since we don't need to load the whole User and Promotion object.

```suggestion
            const promotion = await getCustomRepository(PromotionRepository).hasId(pid);
            const user = await getCustomRepository(UserRepository).hasId(uid);
            if (promotion && user) {
              const savedPromotion = getCustomRepository(UserRepository).create({userid: uid, promotionId: pid})
              const result = await getCustomRepository(SavedPromotionRepository).save(savedpromotion)
            } else {
              // throw some error
            }
            
```

backend/src/main/repository/UserRepository.ts
```diff
@@ -28,13 +29,13 @@ export class UserRepository extends Repository<User> {
...
-    return this.save(user);
+  addSavedPromotion(user: User, promotion: Promotion): Promise<SavedPromotion> {
+    const savedPromotion = new SavedPromotion(user, promotion);
+    return getCustomRepository(SavedPromotionRepository).save(savedPromotion);
+  }
+
+  // remove a single promotion from user's saved promotions
+  removeSavedPromotion(user: User, promotion: Promotion): Promise<DeleteResult> {
+    return getCustomRepository(SavedPromotionRepository).delete({userId: user.id, promotionId: promotion.id});
   }
```
ericrkuo, line 40: Do you think it's better to move these methods into `SavedPromotionRepository`? Because it does not use `UserRepository` methods.

backend/src/main/controller/UserController.ts
```diff
@@ -0,0 +1,185 @@
...
+    getUploaded = async (req: Request, res: Response, next: NextFunction) => {
+        try {
+            // get the id of user from the url
+            const id = await IdValidation.schema.validateAsync(req.params.id, {
+                abortEarly: false,
+            });
+            const userRepository = getCustomRepository(UserRepository);
+            const uploadedpromotions = await userRepository.findOneOrFail(id, {
+                relations: ["uploadedPromotions"]
+            });
```
ericrkuo, line 178: @andrewzulaybar  same for this part, how much detail do we want to show for uploaded promotions.

I was thinking the same reasoning for saved promotions.

I can imagine in the future we allow users to edit their uploaded promotions. Then when they click `edit` on their promotion in the `uploaded promotions page`, then we send a request to get all the necessary details and joins.

backend/src/main/validation/UserUpdateValidation.ts
```diff
@@ -0,0 +1,22 @@
+import Joi, { ObjectSchema } from 'joi';
+
+/**
+ * check the validity of an user, used when we make a POST request to /users
```
ericrkuo, line 4: Do you mean `PATCH` request to /users

ericrkuo (commented):
Very nice work @yunishimura0716 !!!

backend/src/main/constant/Route.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Constants for any routes we use
+ * */
+export enum Route {
+    PROMOTIONS = '/promotions',
+    USER = '/users'
```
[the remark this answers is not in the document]
yunishimura0716: yes! I'll do!

backend/src/main/App.ts
yunishimura0716: thanks! I'll remove it!

backend/src/main/controller/UserController.ts
```diff
@@ -0,0 +1,185 @@
...
+    getSaved = async (req: Request, res: Response, next: NextFunction) => {
+        try {
+            // get the id of user from the url
+            const id = await IdValidation.schema.validateAsync(req.params.id, {
+                abortEarly: false,
+            });
+            const userRepository = getCustomRepository(UserRepository);
+            const savedpromotions = await userRepository.findOneOrFail(id, {
+                relations: ["savedPromotions"]
+            });
```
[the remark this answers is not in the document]
yunishimura0716: Oh, I see, I'll fix it

backend/src/main/repository/UserRepository.ts
yunishimura0716: so, you mean we don't have to define 'addSavedPromotion' and 'removeSavedPromotion' ?

ericrkuo: Like we can define them inside the `SavedPromotionRepository` class rather than `UserREpository` class

yunishimura0716:
I improved the controller method for savedPromotions POST/DELETE. Also, I fixed some merge conflict.

PR #59, opened 2020-11-14
Moved GooglePlacesService to FE 

ld357:
Since we aren't able to save any data (aside from place_id) linked to restaurants indefinitely, I've moved the Google Places API class to the FE instead!

I also added a map to hold restaurant info for the current user's session to prevent any additional unnecessary calls.

NOTE: You may get a warning about the current version of Jest if you try to run the tests using IntelliJ- unfortunately, I had to use the Jest library linked to create-react-app which is the older version that hasn't been tested yet with ts-jest. I wasn't able to update it either since it's in the package.json inside the react-scripts module in node_modules. It still works fine for now- I'll try to see if I can find a way to fix it though!

The information returned from the calls to Google Places Search and Details have modified to match with the algorithm we're currently aiming for: [Link to Interaction Details between Google API and Front/Backend](https://docs.google.com/document/d/1mx9piwHnu8RCeKOoTKDRPSX61MGrNK6vzAfI0qHLCBc/edit?usp=sharing) This is still being modified so we're free to move things around :)


frontend/src/services/GooglePlacesService.ts
```diff
@@ -1,33 +1,34 @@
-import axios, { AxiosError, AxiosResponse } from 'axios';
-import * as dotenv from 'dotenv';
+import axios, { AxiosError, AxiosResponse } from "axios";
+import * as dotenv from "dotenv";
 
 /* eslint-disable  no-console */
 // todo: remove this eslint-disable
 class GooglePlacesService {
+  public currRestaurants: Map<string, Record<string, unknown>>;
```
ld357, line 7: Just a note- I had to use Record instead of object because the linter complained!

backend/jest.config.js
```diff
@@ -1,4 +0,0 @@
-module.exports = {
-  preset: 'ts-jest',
-  testEnvironment: 'node',
-};
```
ericrkuo, line 4: How come delete this from BE?

backend/package.json
```diff
@@ -7,7 +7,6 @@
     "build": "tsc --project ./",
     "prestart": "yarn run build",
     "start": "nodemon src/main/App.ts",
-    "test": "jest",
```
ericrkuo, line 10: Wait don't we still need this to run unit tests in BE?

frontend/src/services/GooglePlacesService.ts
ericrkuo: Just curious, what's the `Record<string, unknown>` for?

frontend/src/services/GooglePlacesService.ts
```diff
@@ -1,33 +1,34 @@
...
+
   constructor() {
     dotenv.config();
+    this.currRestaurants = new Map();
   }
 
-  getGooglePlacesSearch(data: string): Promise<JSON> {
```
ericrkuo, line 11: So will this data string contain both the restaurant name and it's address? Because I see you're taking the first candidate, but would we also need the restaurant specific to the address/location.

frontend/src/services/GooglePlacesService.ts
ericrkuo: Like would we want the value to be an interface we define as `RestaurantDetails`?

frontend/test/services/GooglePlacesService.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      });
+  });
+
+  test("Get details for a place using placeID, should be successful", () => {
+    // placeID can only be retrieved from getGooglePlacesSearch
+    return googlePlacesAPI
+      .getRestaurantDetails("ChIJb0n5cWl3hlQRIbVGYLiTEgE")
+      .then((result: Record<string, unknown>) => {
+        expect(result).toHaveProperty("business_status");
+        expect(result).toHaveProperty("formatted_address");
```
ericrkuo, line 31: Will we also store lat lon in the restaurant details?

ericrkuo (changes requested):

ericrkuo:
Nice job lily!!! Looks really good

backend/jest.config.js
ld357: I assumed you guys wouldn't need jest to test but I can add it back in if needed!

frontend/src/services/GooglePlacesService.ts
ld357: Yup, I didn't add it into my description here but I was planning to address handling the edge case for franchises for this week's task! I'm still figuring out how to best handle this case.

frontend/src/services/GooglePlacesService.ts
ld357: I'm using the Record to manage the current state for the user- so all the restaurants they've clicked on or have looked at so far! We could have an interface- I didn't think of this but it might actually be easier connecting with the restaurant cards

backend/package.json
ld357: See other comment!

frontend/test/services/GooglePlacesService.test.ts
ld357: Yes! It's currently in the geometry field but I like your interface idea so I might do that and extract those details out from geometry

backend/jest.config.js
ericrkuo: Oh wait but if we remove this will we still be able to run tests in BE?

frontend/test/services/GooglePlacesService.test.ts
ericrkuo: Sounds good!

frontend/src/services/GooglePlacesService.ts
ericrkuo: Ooh okie got it!

frontend/test/services/GooglePlacesService.test.ts
```diff
@@ -0,0 +1,120 @@
...
+          googlePlacesAPI.currRestaurants.get("ChIJb0n5cWl3hlQRIbVGYLiTEgE")
+        ).toHaveProperty(
+          "url",
+          "https://maps.google.com/?cid=77286563717231905"
+        );
+        expect(
+          googlePlacesAPI.currRestaurants.get("ChIJb0n5cWl3hlQRIbVGYLiTEgE")
+        ).toHaveProperty("user_ratings_total", 66);
+        expect(
+          googlePlacesAPI.currRestaurants.get("ChIJb0n5cWl3hlQRIbVGYLiTEgE")
```
andrewzulaybar, line 70: Could `googlePlacesAPI.currRestaurants.get("ChIJb0n5cWl3hlQRIbVGYLiTEgE")` be defined as a local variable at the beginning of this `then` block to reduce a bit of the duplication? Same for the following test?

frontend/src/services/GooglePlacesService.ts
```diff
@@ -40,19 +41,30 @@ class GooglePlacesService {
...
     return axios
-      .get('https://maps.googleapis.com/maps/api/place/details/json?', {
+      .get("https://maps.googleapis.com/maps/api/place/details/json?", {
         params: {
           place_id: placeID,
           fields:
-            'url,formatted_phone_number,opening_hours,website,review,photos',
+            "url,formatted_phone_number,opening_hours,website,review,photos,business_status," +
+            "formatted_address,geometry,name,price_level,rating,user_ratings_total",
           key: process.env.GOOGLE_PLACES_API_KEY,
```
andrewzulaybar, line 60: Something I just discovered recently that we could do here is to change this to `process.env.REACT_APP_GOOGLE_PLACES_API_KEY` and add this key to our .env file. I believe `react-scripts` loads these environment variables for us already so we don't also need to list `dotenv` as a dependency!

https://stackoverflow.com/a/56668716

frontend/src/services/GooglePlacesService.ts
andrewzulaybar: Ya I like the idea of having an interface! I've started defining some under `src/types/` so if you wanted to add one for `RestaurantDetails` in there that would be great!

andrewzulaybar (approved):
This looks great! Thanks for getting those tests set up too :)

frontend/src/services/GooglePlacesService.ts
richardyjtian: Yes! This was the way I did it for integrating the google maps api too

richardyjtian (approved):
Everything looks good to me! Great job Lily!

backend/jest.config.js
ld357: You won't be able to if you guys were intending to use ts-jest! Have you been using it for any so far?

frontend/test/services/GooglePlacesService.test.ts
ld357: Yup!! :)

frontend/src/services/GooglePlacesService.ts
ld357: OH WOAH! That's handy 😅  I'll change it to that instead!

backend/jest.config.js
ericrkuo: Hmm wait do you mind explaining which parts of the `GooglePlaceService` were ts-jest when it was in the BE?

frontend/src/services/GooglePlacesService.ts
```diff
@@ -0,0 +1,183 @@
...
+      .get("https://maps.googleapis.com/maps/api/place/details/json?", {
+        params: {
+          place_id: currPlaceID,
+          fields: "formatted_address",
+          key: process.env.REACT_APP_GOOGLE_PLACES_API_KEY,
+        },
+      })
+      .then((response: AxiosResponse) => {
+        const currRestaurantAddress = response?.data?.result?.formatted_address;
+        if (currRestaurantAddress.includes(restaurantLocation)) {
```
ericrkuo, line 90: Hmm how do we know that our restaurant location will match to `currRestaurantAddress`? Using string operations is fragile imo because we can't guarantee that it will always match, there may be a lot of edge cases.

frontend/src/services/GooglePlacesService.ts
```diff
@@ -0,0 +1,183 @@
...
+        if (!Array.isArray(result)) {
+          return Promise.resolve(result);
+        }
+
+        for (let i = 0; i < result.length; i++) {
+          if (result[i]) {
+            return Promise.resolve(responseData[i].place_id);
+          }
+        }
+
```
ericrkuo, line 62: Does google places follow some kind of format where the element in the first array is the one that matches the location + name the best? If it is, I would recommend we just choose the first one at index 0.

ericrkuo: If not, this will be more complicated because of comment mentioned below

frontend/src/services/GooglePlacesService.ts
```diff
@@ -22,13 +27,40 @@ class GooglePlacesService {
...
-            key: process.env.GOOGLE_PLACES_API_KEY,
+            key: process.env.REACT_APP_GOOGLE_PLACES_API_KEY,
           },
         }
       )
       .then((response: AxiosResponse) => {
-        const resultPlaceID: string = response.data.candidates?.[0].place_id;
-        return Promise.resolve(resultPlaceID);
+        if (response.data.candidates.length === 1) {
+          return Promise.resolve(response.data.candidates?.[0].place_id);
```
andrewzulaybar, line 36: Does `response.data.candidates[0].place_id` (without the optional chaining) not work here? I'm assuming `response.data.candidates` is defined at this point but I could be wrong!

frontend/src/services/GooglePlacesService.ts
andrewzulaybar: Ya, this is a good suggestion. @ld357 If this isn't possible, then I think we can go ahead with this implementation as is but I think I'll create a ticket to look into something like https://developers.google.com/places/web-service/autocomplete but we could probably discuss this more on Saturday!

frontend/src/services/GooglePlacesService.ts
ld357: Yeah that's true, I see where you're coming from. I couldn't think of another way to verify if the current restaurant is the one that the user is searching for though- and also with correspondence of the fact that we're storing the user-submitted locations in the DB, I think _for now until we figure out a better way to handle both situations_, we can add extra notes in the promotion submission page that tells the user to make sure they're inserting the address correctly (or ask for particular fields like Address Line 1, Address Line 2, City, Province, Postal Code)?

frontend/src/services/GooglePlacesService.ts
ld357: Ahhh yeah I was hoping to do that too when I was researching into it, but it isn't possible 😢   Right now I'm doing a Places Search - Find Place request (which takes in a name, address, or phone number) and then a Places Details request to check the location match. I decided to take in the name since I feel like it's less likely to be incorrect compared to an address.

There is another one called "Places API Text Search Service" which allows searches like "tacos in Vancouver" or "123 E Broadway" which will return a list of places that seem to match. I don't think it's the best idea though because it's more generic and relies heavily on the user's submitted address.

frontend/src/services/GooglePlacesService.ts
ld357: Sorry 😅  what are you asking about in particular? I'm checking if the length here is 1 because that means there weren't any other restaurants found with similar names (either because it's a franchise or just multiple restaurants with that name)

backend/jest.config.js
ld357: The old tests for GooglePlacesService were!

ericrkuo: I think we need to keep this for BE because we use it here https://github.com/ubclaunchpad/foodies/blob/master/backend/jest.config.js. We mainly need it because of this: https://github.com/ubclaunchpad/foodies/pull/76/files

ld357: All good! I'll add it back in :)

frontend/src/services/GooglePlacesService.ts
```diff
@@ -0,0 +1,183 @@
...
+            key: process.env.REACT_APP_GOOGLE_PLACES_API_KEY,
+          },
+        }
+      )
+      .then((response: AxiosResponse) => {
+        if (response.data.candidates.length === 1) {
+          return Promise.resolve(response.data.candidates[0].place_id);
+        }
+        const promises: Array<Promise<boolean>> = [];
+        responseData = response.data.candidates;
```
ericrkuo, line 39: Also, can we make the `findplacefromtext` return the address as well? I think it can return something called `formatted_address` which may be the same as the `details` endpoint. Then we don't need to make another http request for each potential candidate.
image

frontend/src/services/GooglePlacesService.ts
```diff
@@ -0,0 +1,183 @@
...
+// todo: remove this eslint-disable
+class GooglePlacesService {
+  public currRestaurants: Map<string, RestaurantDetails>;
+
+  constructor() {
+    dotenv.config();
+    this.currRestaurants = new Map();
+  }
+
+  getRestaurantPlaceID(
```
ericrkuo, line 15: Will this function also need to return the `lat lon` values for the backend to store?

ericrkuo (commented):
Looks great! Just some general comments hoping to clarify

backend/package.json
```diff
@@ -36,7 +36,7 @@
   },
   "devDependencies": {
     "@types/express": "^4.17.8",
-    "@types/jest": "^26.0.15",
+    "@types/jest": "^26.0.19",
```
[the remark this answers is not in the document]
ld357: Since I re-installed it, I updated as well!

frontend/src/services/GooglePlacesService.ts
```diff
@@ -0,0 +1,183 @@
...
+        const promises: Array<Promise<boolean>> = [];
+        responseData = response.data.candidates;
+
+        for (const currPlace of response.data.candidates) {
+          promises.push(
+            this.isCorrectRestaurantLocation(
+              restaurantLocation,
+              currPlace.place_id
+            )
+          );
```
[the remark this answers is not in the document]
ld357: The rate limit is 100 requests per second!

frontend/src/services/GooglePlacesService.ts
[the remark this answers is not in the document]
ld357: Yup! I forgot if we brought it up here but we should consider adding a documentation library later on? @andrewzulaybar

backend/package.json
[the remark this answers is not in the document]
ericrkuo: Ahh okay got it, looks like it doesn't break any existing behaviour, thanks for updating the dependencies!

frontend/src/services/GooglePlacesService.ts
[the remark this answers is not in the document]
ericrkuo: Wauw that's very good! Makes sense thanks!

frontend/src/services/GooglePlacesService.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        return Promise.all(promises);
+      })
+      .then((result: Array<boolean>) => {
+        if (!Array.isArray(result)) {
+          return Promise.resolve(result);
+        }
```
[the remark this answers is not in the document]
ld357: Woops lemme fix this!

frontend/src/services/GooglePlacesService.ts
ld357: Nope! We call getRestaurantDetails after receiving the placeID

ericrkuo (approved):

frontend/src/services/GooglePlacesService.ts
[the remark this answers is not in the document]
andrewzulaybar: @ld357 @ericrkuo Oh yeah I think I've just been adding comments here and there but no formal library for documentation. I'll set up an issue for adding https://tsdoc.org/

PR #63, opened 2020-11-17
Added options and theme

richardyjtian:
I've created an options and theme component to use in your components. Options should be mostly complete except for the links, and please feel free to add to the theme whichever settings you feel we may frequently use.


richardyjtian:
Actually, I'm not sure this is the best way to reference the font. Maybe let's just use the default fontfamily and fontsize for simplicity?

PR #67, opened 2020-11-19
Add filtering/sorting for dropdowns on Home page

andrewzulaybar:
**Issues:**
n/a

**Summary:**
I added the ability for the dropdowns on the Home page to dynamically filter the current list of promotions.

**Changes:**
- Created `PromotionsListProvider` which uses React's `useReducer` and `Context` APIs to achieve globally injectable state management (just import and then use `usePromotionsList` whenever needed on the Home page)
- Created the `promotions` service, which is just a file with a list of TypeScript functions to handle API calls and/or operate on objects of type `Promotion`. Some functions still need to be finished, just waiting on #59 to get merged in
- Changed the structure for defining dropdowns so that we could specify an `onClick` callback function for each dropdown option
- Created types that would act as options for the Dropdown menu. This should eventually be taken from the backend, just waiting on #52 to get merged in

**Quick demo:**
https://www.loom.com/share/b83d68fa434c45d8884d6486918be1f0


frontend/src/types/promotion.ts
```diff
@@ -0,0 +1,82 @@
...
+  id: string;
+  discountValue: number;
+  type: string;
+}
+
+export interface PromotionImage {
+  src: string;
+}
+
+export enum Category {
```
richardyjtian, line 35: I think there may be some miscommunication on the meaning of "Category". Backend and design seem to have this as promotion category (ie. bogo, happy hour, etc), but I agree with you that it makes more sense that category refers to restaurant category (especially when "Discount Type" is itself a thing)

richardyjtian (approved):

richardyjtian:
@andrewzulaybar How do you feel about me merging and integrating the "My Favourites" page into the home page? I've thought about it and it seems a little awkward for favourited promotions to be its own page because it requires most of the functionality of the home page (dropdown menu, popup restaurant card, maybe map even) with the addition of a "sort by: recently added" and "filter by: favourited". I can make it so that these two options appear in the home screen once a user's logged in.

frontend/src/types/promotion.ts
ericrkuo: Ah I see okay so this is a new DB schema, this makes more sense I agree.

So just to double-check, a promotions category will go more specific into the cuisine type. And then the `Discount` entity will tell us the discount type and discount category? Discount type is for specific numeric units like percentage, dollar amount etc. But discount category will still exist to be Bogo, happy hour?

One thing I'm thinking about is now the complexity of preventing users from choosing a category that does not belong in the cuisine type. For example, a user could create a promotion with a category "curry", but then call that cuisine type "Mexican"

ericrkuo: This will take some time to refactor in the BE too unfortunately

ericrkuo:
> @andrewzulaybar How do you feel about me merging and integrating the "My Favourites" page into the home page? I've thought about it and it seems a little awkward for favourited promotions to be its own page because it requires most of the functionality of the home page (dropdown menu, popup restaurant card, maybe map even) with the addition of a "sort by: recently added" and "filter by: favourited". I can make it so that these two options appear in the home screen once a user's logged in.

I like this idea @richardyjtian, but note that there will be an extra field returned since we store the time the promotion was saved for the user. In the future, if we add more columns/data for a user's saved promotions, then it may need to be separated out into its own page.

If we were to do this approach, the backend could also modify the saved promotions endpoint to just return a list of the promotions by ID rather than the whole promotion object, that way you can just do an intersection with the existing promotions you got back from GET promotions

frontend/src/types/promotion.ts
richardyjtian: I'm not sure that a discount category is necessary to be honest. I think the current schema should be good as long as the "category" field is understood to be restaurant category. Also, I think examples of restaurant category would be like the ones below (ie. fast food, fast casual, casual dining, contemporary casual, cafe)

ericrkuo: But then how would a user filter for promotions that are specifically happy hour, BOGO, seasonal etc

ericrkuo: In terms of restaurant categories, I can also imagine this being supported by our search query functionality. The user can just search up fast food, and we can search inside the promotion name and description for data like this

richardyjtian: My thinking was that users wouldn't see much use in filtering by discount category, but I could be wrong. It's probably best to keep discount category though and add it to the discount table if it's not too much work @ericrkuo

andrewzulaybar: @richardyjtian @ericrkuo Ya, I think it got lost in translation somewhere... So I think there are two types of categories: Restaurant Categories ("bubble tea", "dessert", "fast food") and Promotion Categories (BOGO, Happy Hour). I had a discussion with Wendy and Elina last week about it and they gave me these categories, but then they didn't account for the promotion categories. Now I'm looking at the hi-fi designs and they're both missing so yeah I'll need to clear this up!

andrewzulaybar: Screen Shot 2020-11-18 at 9 07 02 PM
Screen Shot 2020-11-18 at 9 07 27 PM

@richardyjtian @ericrkuo What do you guys think about this?

ericrkuo: > My thinking was that users wouldn't see much use in filtering by discount category, but I could be wrong. It's probably best to keep discount category though and add it to the discount table if it's not too much work @ericrkuo

Ah yea I see what you mean Richard that makes sense. Like something more useful would be to search a quantifiable filter, like promotions with greater than 10$ off, or greater than 50% off. rather than to filter by %, $, etc.

@andrewzulaybar Elina mentioned to me that the original reasoning for not including restaurant categories was because this could be supported by the search functionality. But then one could also argue, Cuisine Type could also be supported by the search functionality.

ericrkuo: Oops i forgot to refresh before sending my message, but yes I like Elina's idea. I also like richards idea with the discount type

andrewzulaybar:
> @andrewzulaybar How do you feel about me merging and integrating the "My Favourites" page into the home page? I've thought about it and it seems a little awkward for favourited promotions to be its own page because it requires most of the functionality of the home page (dropdown menu, popup restaurant card, maybe map even) with the addition of a "sort by: recently added" and "filter by: favourited". I can make it so that these two options appear in the home screen once a user's logged in.

Ah, I see what you mean because they're pretty similar. I think the intention behind having a separate page is similar to Yelp's functionality where I think they have their bookmarks on a separate page. I think this way sets it up for us in the future if we wanted to expand functionality on a user's favourite promotions and so that the Home page isn't more cluttered. But I see your point and I think that could be a good idea. If we want to propose this, we should discuss it Wendy and Elina and the rest of the team in the slack channel

frontend/src/types/promotion.ts
ericrkuo: actually now i am confused, what happened to Promotion Category LOL

andrewzulaybar: > Ah yea I see what you mean Richard that makes sense. Like something more useful would be to search a quantifiable filter, like promotions with greater than 10$ off, or greater than 50% off. rather than to filter by %, $, etc. @andrewzulaybar would you mind creating an issue for this if you're oaky with the idea?

By search by a quantifiable filter do you mean using the search bar?

andrewzulaybar: > actually now i am confused, what happened to Promotion Category LOL is there a filter for that still?

Ya I'm a bit confused too haha let me check in with Elina

andrewzulaybar: Screen Shot 2020-11-18 at 9 20 06 PM

@ericrkuo I think they're here now? They've combined it with the "Discount Type" but Elina's asking me if "Promotion Type" would be a better name for this dropdown

ericrkuo: Hmmm
looking at the figma it seems Promotion Category and Discount Type are merged together. That could work imo
image

ericrkuo: Oops didnt refresh again sry

PR #72, opened 2020-11-20
Fixed bug with DTO date

ericrkuo:
DTO will already create a Date object automatically for us, there was no difference between `new Date(promotionDTO.expirationDate)` vs `promotionDTO.expirationDate`


yunishimura0716 (approved):
Looks good!

ld357 (approved):
AMAZING 💯  !!!!!

Issue #51, opened 2020-11-11
Backend needs Schedule entity


ericrkuo:
Need to discuss how we want to format `Promotion_Time` entity in the backend and communicate that format with frontend.

Requirements
- need to support recurring and non-recurring promotions
- need to be able to support all days of the week at specific times
- recurring needs to be able to handle weekly, monthly recurrences (yearly is extreme)
- non-recurring needs to handle single day promotion, or all days of a specific week, or just multiple days that aren't recurring
- need to be able to filter promotions by day (maybe multiple days??)

Discussions:
https://docs.google.com/document/d/131JfzQXb472CyOL2yMSqbfuDyGI13UwQgnh0kowzPuk/edit?disco=AAAAKYqRHew
- with this approach above ^
  - we need to think more about unique entries in our DB (for e.g. if we say promotionId and day of the week is unique, we can only have one promotion schedule (recurring or non-recurring) for each day of the week, so upper bound is 7). But this won't allow us to say Monday Nov X schedule is 1-3pm, Monday Nov X+7 Schedule is 10am-5pm for example. Depends on our use cases

![image](https://user-images.githubusercontent.com/49849754/98756965-66a7ef80-2380-11eb-8ef8-318cc8d47e21.png)

PR #75, opened 2020-11-21
51 - Backend needs Schedule entity

ericrkuo:
https://github.com/ubclaunchpad/foodies/issues/51

- Added new `Schedule` entity.
  - The schedule represents the days of the week and time a promotion is active (the specific dates a promotion is valid is represented by `startDate` (explained below) and `expirationDate` (existing behaviour)
  - Each schedule has a `startTime, endTime, dayOfWeek, and isRecurring` field.
  - As you can see, we only allow schedules to repeat weekly or not. There is no biweekly, monthly, yearly etc. This was mainly the decision to simplify stuff on our end, and until we have a higher demand for more flexible scheduling, I think this design will suit our needs!
- Each promotion must have at least one schedule and can have at most seven
   - at most seven schedules because one schedule per day of the week
- Added `ScheduleDTO` and validation for the schedule
- **important:" Promotion has a new column called `startDate`
  - represents when a promotion starts (different from `date_added`)
- modified `GET promotions` endpoint to return schedules of all promotions
- modified `POST promotions` endpoint to take in a schedule and validate it
- modified `GET promotion` to return schedules of a single promotion
- added `ScheduleRepository`
- made sure cascade delete works (when delete promotion will delete corresponding schedules)
- updated sample data and added new unit tests

This is what a single promotion looks like now
image


backend/src/main/controller/PromotionController.ts
```diff
@@ -89,14 +91,27 @@ export class PromotionController {
...
+      const schedules = promotionDTO.schedules.map(
+        (scheduleDTO: ScheduleDTO) => {
+          return new Schedule(
+            scheduleDTO.startTime,
+            scheduleDTO.endTime,
+            scheduleDTO.dayOfWeek,
+            scheduleDTO.isRecurring
+          );
+        }
+      );
```
ericrkuo, line 103: One thing to note is that the incoming request body will validate **all** the discounts and schedules automatically as well, it's very powerful 😲 

This is just a couple of lines above, but note how I didn't need to explicitly mention, validate the schedule part of the `request.body` or validate the discount part of the `request.body`
https://github.com/ubclaunchpad/foodies/blob/b42162ce71da7c85e879017ac4d1f3aa98bcbbcc/backend/src/main/controller/PromotionController.ts#L82-L85

This is an example POST request, and all this is the format of a PromotionDTO. Also take a look at `PromotionValidation.ts` to see how we are validating `Discount`'s and `Schedules` as well if you are curious.

image

backend/src/main/validation/ScheduleValidation.ts
```diff
@@ -0,0 +1,80 @@
...
+      .regex(timeRegex)
+      .messages(startTimeMessages)
+      .required()
+      .custom(customTimeValidator), // only use time validator for startTime so we do not get duplicate start/end time error messages
+    endTime: Joi.string().regex(timeRegex).messages(endTimeMessages).required(),
+    isRecurring: Joi.boolean().required(),
+    dayOfWeek: Joi.string()
+      .valid(...Object.values(Day))
+      .required(),
+  }).required();
```
ericrkuo, line 72: This class is more complicated because Joi does not support validating time. They have a module library called `joi-date` but it is not compatible with typescript.

But to summarize what is going on here, we validate a start and end time using regex. And we use our own custom validator to verify the start time is less than the end time. 

We also create our own custom error message if the string does not match the regex, this is using `startTimeMessages` and `endTimeMessages`. How this works is that when the regex does not match, Joi uses the code `string.pattern.base` to signal that the string doesn't match the regex. All we have to do is register a custom error message associated with that code

ericrkuo (commented):
Let me know if anything can be improved, open to suggestions and feedback. This PR is actually relatively small compared to other ones 😅 A lot of the files are just mainly unit/integration tests, updating sample data, cleaning up old code irrelevant to schedules.

backend/src/main/validation/PromotionValidation.ts
```diff
@@ -20,17 +31,30 @@ export class PromotionValidation {
...
+     * startDate is NOT required. If is null/undefined, then default to current date.
+     * However, Promotion.startDate cannot be null or undefined
+     * * note incoming format is a string, and Joi will automatically convert to type Date
+     * * startDate cannot be larger than expirationDate
+     * */
+    startDate: Joi.date()
+      .max(Joi.ref('expirationDate'))
+      .empty(null)
+      .default(new Date()),
+    /** note incoming format is a string, and Joi will automatically convert to type Date */
```
ericrkuo, line 44: @andrewzulaybar just moving what we talked about here so everyone can input ideas.

Right now we are thinking when the user is choosing the start date of a promotion, if they indicate somehow that the promotion is active (e.g. ticking a checkbox rather than selecting a start date), then FE will leave startDate as null/undefined when they send the POST request to promotions. 

This will set the start date of the promotion automatically to the current date.

The reasoning for this is because say we want to show in the promotions card whether a promotion is active or whether it has started or not. We can then use date comparison tools, e.g. if `promotion.startDate <= <today's date>` then we will show that the promotion is active, otherwise we would show the promotions start date.

backend/src/main/validation/PromotionValidation.ts
```diff
@@ -12,6 +14,16 @@ export class PromotionValidation {
...
+    /**
+     * A promotion must have at least one schedule and can have at most seven (each representing one day of the week)
+     * Will validate each item in array with ScheduleValidation's schema
+     * */
+    schedules: Joi.array()
+      .unique('dayOfWeek', { ignoreUndefined: true }) // ignore undefined so that in case object does not contain dayOfWeek at all
+      .min(1)
+      .max(7)
+      .items(ScheduleValidation.schema)
+      .required(),
```
ericrkuo, line 26: Let me know if this is okay! To me, I did not see a point for a user having more than 7 schedules.

Also, the validation message is not user friendly when we are checking for duplicate dayOfweek's for the array of schedules. It says something liek `schedules[1] contains a duplicate value`. In the future I can imagine something like this:

```
fun validate(target: any) {
 const errors = ... // get all validation errors from Joi
 errors.addAll(PromotionDTO.validateMoreComplicatedStuff(target as PromotionDTO))
 return errors;
}
```

That way `validateMoreComplicatedStuff` can contain custom logic for more complicated validations and has more user-friendly error messages. But at this point, the purpose of our validation is just to make sure we don't use DB transactions to tell us whether anything is ill-formatted. Better user-friendly messages should be another issue.

backend/src/main/validation/PromotionValidation.ts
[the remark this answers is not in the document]
yunishimura0716: Yeah, this is good point! We need to discuss more about some validations on FE or BE.

yunishimura0716:
I'm just thinking that what if the user want to promote a deal during specific date (E.G. the deal is valid from 12/10 ~ 12/25).

backend/src/main/validation/ScheduleValidation.ts
yunishimura0716: Just want to clarify, so regex checks if the input string match with the patterned string, right?

ericrkuo:
> I'm just thinking that what if the user want to promote a deal during specific date (E.G. the deal is valid from 12/10 ~ 12/25).

Ah yea so then we would set promotion `startDate` as 12/10, promotion `expirationDate` as 12/25. Then in DB we would store 7 `Schedule` entities, each representing one day of the week with `isRecurring = true`. 

Then I think FE would need to be able to display this information correctly.

backend/src/main/validation/ScheduleValidation.ts
ericrkuo: Yup that's correct!

yunishimura0716:
> > I'm just thinking that what if the user want to promote a deal during specific date (E.G. the deal is valid from 12/10 ~ 12/25).
> 
> Ah yea so then we would set promotion `startDate` as 12/10, promotion `expirationDate` as 12/25. Then in DB we would store 7 `Schedule` entities, each representing one day of the week with `isRecurring = true`.
> 
> Then I think FE would need to be able to display this information correctly.

What is value of `startTime` and `endTime` at the case?

ericrkuo:
> > > I'm just thinking that what if the user want to promote a deal during specific date (E.G. the deal is valid from 12/10 ~ 12/25).
> > 
> > Ah yea so then we would set promotion `startDate` as 12/10, promotion `expirationDate` as 12/25. Then in DB we would store 7 `Schedule` entities, each representing one day of the week with `isRecurring = true`.
> > 
> > Then I think FE would need to be able to display this information correctly.
> 
> What is value of `startTime` and `endTime` at the case?

Oh the user will specify value of start time and end time in FE, so it's up to them to specify it.

yunishimura0716 (approved):
Thank you for your great work!

backend/src/main/validation/PromotionValidation.ts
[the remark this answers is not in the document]
ld357: I think there's a lot of diff validations we can do.. Maybe one validation is check that the endDate has the same weekday as one of the schedules- so if it ends on Monday, Dec 28 then one of the schedules should include the Monday or else it's invalid

backend/src/main/controller/PromotionController.ts
ld357: Is the time automatically PST/PDT?

backend/src/main/validation/PromotionValidation.ts
```diff
@@ -20,17 +32,30 @@ export class PromotionValidation {
...
-    expirationDate: Joi.date().required(), // note incoming format is a string, and Joi will automatically convert to type Date
+    /**
+     * startDate is NOT required. If is null/undefined, then default to current date.
+     * However, Promotion.startDate cannot be null or undefined
+     * * note incoming format is a string, and Joi will automatically convert to type Date
+     * * startDate cannot be larger than expirationDate
+     * */
+    startDate: Joi.date()
+      .max(Joi.ref('expirationDate'))
+      .empty(null)
```
ld357, line 43: Just curious but what does `.empty `do?

backend/src/main/validation/ScheduleValidation.ts
```diff
@@ -0,0 +1,80 @@
...
+  try {
+    const timeParts: string[] = time.split(':');
+    return Number(timeParts[0]) * 60 + Number(timeParts[1]);
+  } catch (e) {
+    return null;
+  }
+}
+
+/**
+ * Checks the validity of a Discount
```
ld357, line 57: Do you mean Schedule here?

backend/src/main/validation/ScheduleValidation.ts
```diff
@@ -0,0 +1,80 @@
...
+ * */
+export class ScheduleValidation {
+  // note: since schema is marked as required, this means PromotionValidation's schedules must contain at least one schedule
+  static schema = Joi.object({
+    startTime: Joi.string()
+      .regex(timeRegex)
+      .messages(startTimeMessages)
+      .required()
+      .custom(customTimeValidator), // only use time validator for startTime so we do not get duplicate start/end time error messages
+    endTime: Joi.string().regex(timeRegex).messages(endTimeMessages).required(),
```
ld357, line 67: How does the endTimeMessage get thrown since we only use the validator with startTime?

ld357 (commented):
Looks awesome!! 👍🏻  Just a few questions 😄

backend/src/test/repository/EntityIntegration.test.ts
```diff
@@ -294,4 +306,43 @@ describe('Integration tests for all entities', function () {
       fail('Should not have failed: ' + e);
     }
   });
+
+  test('Cascade delete - deleting a promotion will delete its schedules', async () => {
+    const promotion = promotions_sample[0];
+    assert(promotion.discount !== null);
+
+    // persist into db
+    await userRepository.save(promotion.user);
```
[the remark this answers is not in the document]
ericrkuo: Oh my only reason was because I don't expect the persisting to fail, this part is essentially saving the data so that it's there for our integration test to use.

backend/src/main/validation/ScheduleValidation.ts
ericrkuo: Ah yup lemme go fix that thanks!

backend/src/main/validation/ScheduleValidation.ts
ericrkuo: Oh because we only need to use the `customValidator` once, either with start time or end time. Otherwise if startTime > endTime, then we would get duplicate error messages

backend/src/main/validation/PromotionValidation.ts
```diff
@@ -12,6 +14,16 @@ export class PromotionValidation {
     userId: IdValidation.schema,
     placeId: Joi.string().required(),
     discount: DiscountValidation.schema,
+    /**
+     * A promotion must have at least one schedule and can have at most seven (each representing one day of the week)
```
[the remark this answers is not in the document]
ericrkuo: Do you mean non-recurring? How come two schedules instead of one?

backend/src/main/entity/Schedule.ts
```diff
@@ -24,4 +44,29 @@ export class Schedule {
   })
   @JoinColumn()
   promotion: Promotion;
+
+  /**
+   * Represents 24 hour time format (hh:mm:ss or hh:mm:ss).
+   * Does not support timezones yet
```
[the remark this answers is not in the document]
ericrkuo: I think so, we can worry about globalization afterwards!

backend/src/main/validation/PromotionValidation.ts
ericrkuo: Oh essentially, if the startDate is either undefined or null, then we default to the current date. `empty(null)` just says we consider `null` as undefined, so then Joi would default that to the current date. 

image

backend/src/main/validation/PromotionValidation.ts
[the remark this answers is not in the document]
ld357: Oooh!! Yes my bad I meant non-recurring 😅  but I was thinking for promotions with a startDate and endDate, we would need exactly X schedules based on X number of days in between start and end? This would be a validation thing though I guess?

I was originally thinking of 2 in the case that it starts Dec 27 and ends Dec 28! But also for the case with at least one schedule- would that be either a promotion that has no end date or the start = end?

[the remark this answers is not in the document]
ld357: Not sure if you did this already but maybe we should create an issue of all the validations we can do?

backend/src/main/validation/PromotionValidation.ts
ld357: Ooooh woah that's cool- thanks! :)

backend/src/main/validation/ScheduleValidation.ts
ld357: Yup that makes sense but how does the endTimeMessage get thrown if we registered it only with the start one?

ericrkuo: Ohhh wait sorry I think I misunderstood what you meant. `endTimeMessage` is for invalid regex format, so that's automatically registered since we specified `messages(endTimeMessages)` and because here, Joi looks for the key `string.pattern.base`
image

backend/src/main/validation/PromotionValidation.ts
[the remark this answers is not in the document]
ericrkuo: Hmm, but startDate and expirationDatecould be months apart and could have a lot of repetition. Note that each schedule doesn't have a calendar date associated with it, only a time and weekday.

If it's Dec 27 and Dec 28, we would have one promotion with that startDate and expirationDate. And then followed by two schedules, one for Sunday and another for Monday, both non-recurring. 

> But also for the case with at least one schedule- would that be either a promotion that has no end date or the start = end?

Do you mean `for the case with exactly one schedule`? That would be a promotion that's only valid on one day.

> Not sure if you did this already but maybe we should create an issue of all the validations we can do?

Do you mean for the frontend or backend?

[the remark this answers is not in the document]
ld357: Ahh that's true and I see that makes sense! Probably on FE since we should do all our validations before storing it in the db and whatnot

backend/src/main/validation/ScheduleValidation.ts
ld357: Gotcha! This what was I was asking- thank you!

backend/src/main/controller/PromotionController.ts
ericrkuo: yup!

backend/src/main/validation/PromotionValidation.ts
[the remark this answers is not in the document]
ericrkuo: yeah this is some good validations to have. Let's create an issue for missing validations that we can tackle all at once. Then we can separate which can be validated front end or backend or both

backend/src/main/validation/PromotionValidation.ts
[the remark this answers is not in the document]
ericrkuo: Yup I agree!

ld357 (approved):

PR #76, opened 2020-11-21
Optimized tests

ericrkuo:
Optimized our integration/unit tests to run faster and fixed bug where we cannot run all tests at once

_Optimizing test_

- The trick here is the `BaseRepositoryTest.ts`. Before for every test that interacts with the `foodies_test` database, we were creating a new connection **each time** a test began. But this was very slow.
- Now, each test suite creates only **one** connection for all the tests running. And before each test begins, we manually clear the database by executing SQL DELETE statements to delete from all the tables, rather than dropping the connection and restarting a new connection (which would drop all the tables).

Our longest test class is `EntityIntegrationTest.ts`. This is how long it took before:
image
This is how long it takes now, we have optimized the tests to be almost **8x faster**
image

_Running all tests at once_
- see the README.md for instructions now
- we also fixed the bug, so now running `yarn run test` will succeed

Now all our tests take about 1 second
image

Before all tests took this long
image

Credits to https://dev.to/caiulucas/tests-with-jest-and-typeorm-4j1l for the idea!

### DIFF
--- a/backend/jest.config.js
+++ b/backend/jest.config.js
@@ -1,4 +1,6 @@
 module.exports = {
   preset: 'ts-jest',
   testEnvironment: 'node',
+  maxWorkers: 1,
+  testMatch: ['**/__tests__/**/*.ts?(x)', '**/?(*.)+(spec|test).ts?(x)'],
 };
--- a/backend/src/test/repository/BaseRepositoryTest.ts
+++ b/backend/src/test/repository/BaseRepositoryTest.ts
@@ -1,9 +1,9 @@
-import { Connection, createConnection } from 'typeorm';
+import { createConnection, EntityMetadata, getConnection } from 'typeorm';
 
 // use this class whenever we want to write unit/integration tests that interact with foodies_test database.
-export class BaseRepositoryTest {
-  static establishTestConnection(): Promise<Connection> {
-    return createConnection({
+const connection = {
+  async create(): Promise<void> {
+    await createConnection({
       type: 'postgres',
       host: 'localhost',
       port: 5432,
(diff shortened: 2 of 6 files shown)


backend/src/test/repository/BaseRepositoryTest.ts
```diff
@@ -1,9 +1,9 @@
-import { Connection, createConnection } from 'typeorm';
+import { createConnection, EntityMetadata, getConnection } from 'typeorm';
 
 // use this class whenever we want to write unit/integration tests that interact with foodies_test database.
-export class BaseRepositoryTest {
-  static establishTestConnection(): Promise<Connection> {
-    return createConnection({
+const connection = {
+  async create(): Promise<void> {
+    await createConnection({
```
ericrkuo, line 6: This is just a connection object with three methods
```
create, close, clear
```

`close` calls `clear` beforehand

backend/src/test/repository/DiscountRepository.test.ts
```diff
@@ -1,20 +1,23 @@
...
+    await connection.close();
   });
 
-  afterEach(() => {
-    const conn = getConnection();
-    return conn.close();
+  beforeEach(async () => {
+    await connection.clear();
+    discountRepository = getCustomRepository(DiscountRepository);
   });
```
ericrkuo, line 21: So before we run the whole test class, we create a connection, before each test we clear the tables and set the repositories, after all tests run, we clear the tables and close the connection.

**Note** we need to clear the tables after all tests run so that no data is left behind in foodies_test

ericrkuo (commented):
Lmk what you think!

backend/jest.config.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
   preset: 'ts-jest',
   testEnvironment: 'node',
+  maxWorkers: 1,
+  testMatch: ['**/__tests__/**/*.ts?(x)', '**/?(*.)+(spec|test).ts?(x)'],
```
ericrkuo, line 5: The reason why it was not working before, was because jest would execute all tests in parallel, but we cannot afford to do this for our tests that interact with the database.

It was also not working because jest would run the `.js` test files for some reason, but we only want it to run `.ts` test files.

yunishimura0716:
Nice work, @ericrkuo . I hope this can fix the error occurred in test env.

yunishimura0716 (approved):

ld357 (approved):

Issue #73, opened 2020-11-20
Backend: Filter promotions by discount type and discount value


ericrkuo:
As a user, I want to find all promotions that may be greater or equal to X$ or X% off. Or I may want to find all promotions that are $ based or % based.

Implementation details:
Will require modification to GET promotions endpoint and will also need modification to PromotionQueryValidation

PR #77, opened 2020-11-21
73 - Filter promotions by discount type and discount value

ericrkuo:
https://github.com/ubclaunchpad/foodies/issues/73

Existing behaviour
- We could find all promotions with discount type % or $

New behaviour
- We can now find all promotions with discount type % or $ **and** with a discount value greater or equal than the specified value
  - e.g. we can find all promotions with discount over $3 or discount greater than 10%

Approach:
- added to the PromotionQueryDTO new field `discountValue`
- added to the PromotionQueryValidation that if `discountValue` is present, `discountType` must be present as well.
- `discountValue` must be a number (since we send it as a request query, it will be formatted as a string, so it will automatically
-  be convereted to a number). Also if `discountValue` has more than 2 decimal places, will automatically be rounded to 2 decimal places

Other stuff
- Our tests shouldn't use sample data from `Data.ts`. Because say if test 1 modifies `users_sample[0]`. Then if test 2 uses that user, the user would have been modified. Therefore, created Factory classes to generate these entities


backend/src/test/factory/DiscountFactory.ts
```diff
@@ -0,0 +1,11 @@
+import { DiscountType } from '../../main/data/DiscountType';
+import { Discount } from '../../main/entity/Discount';
+
+export class DiscountFactory {
+  generate(discountType?: DiscountType, discountValue?: number): Discount {
+    return new Discount(
+      discountType ?? DiscountType.AMOUNT,
```
ld357, line 7: What does the `??` notation mean? Also wondering why we set it to 188.08 for discountValue?

backend/src/test/factory/DiscountFactory.ts
```diff
@@ -0,0 +1,11 @@
...
+      discountValue ?? 188.08
```
ld357, line 8: By default for this here and for the other factory classes, we're setting defaults? So 188.08 would be the discountValue if the parameter is null/undefined?

ld357: How come we're doing it in certain places too? 😅  Sorry, not too familiar with this!

ld357 (commented):
Looks super clean and nice 310 incorporation 🥳 🙌🏻

backend/src/test/factory/DiscountFactory.ts
ericrkuo: Yup those are default values since the parameters we defined inside `generate` are optional because maybe in testing we don't care about setting certain fields and we leave it to the factory to randomly generate this. 

But if you take a look at `PromotionFactory` you'll see that `user, schedules, discount` are not optional values, because those are mandatory fields we want to enforce when testing

ericrkuo: However, if we wanted to automatically add `user, schedules, discount` when we use the Factory method, we could create another function like this:

```
export class PromotionFactory {
  generateWithUserDiscountAndSchedule( <other parameters>
  ): Promotion {
    return new Promotion(
      new UserFactory().generate(),
      new DiscountFactory().generate(),
      [new ScheduleFactory().generate()]
      ... <other parameters>
    );
  }
}
```

ld357: I see!! Is there a harm to keeping the optional values as null/undefined? I'm just thinking if it might be confusing when we're checking test results or debugging!

ericrkuo: Yeah because we have some non-nullable columns, so we can't persist them if they're undefined

ericrkuo: But yeah it should be pretty okay like I don't see us writing too many tests that concern some of the undefined fields in the factory methods. And if it is confusing for the person who's debugging or writing the tests, what's nice is that you can provide your own values to the factory methods if you want.

ld357: Hmm, so even though we're only providing these default values for optional variables, you're saying that some of these can't be null/undefined?

ericrkuo: Yeah like these optional parameters are non-nullable in the DB. I don't think we can persist undefined, but regardless like I think generating these random values won't do harm because if it's confusing we can just override them!
image

ld357: Oooooh okay! As long as we remember with our tests in the future, but sounds good!

ld357 (approved):

yunishimura0716:
Nice work, Eric!! I have a basic question. What is `factory` directory in test directory?

ericrkuo:
> Nice work, Eric!! I have a basic question. What is `factory` directory in test directory?

Ah yeah so the factory directory is used to construct all kinds of entities (like User, Promotion, Schedule). The factory directory has a whole bunch of factory classes. Ideally, we want to use these factory classes for our tests, rather than relying on the sample data (e.g. users_sample, promotions_sample)

yunishimura0716:
I see!! That is nice idea. Thank you!!

yunishimura0716 (approved):

Issue #26, opened 2020-11-01
Add sidebar to "My Account" page


andrewzulaybar:
## 🔨 Task

Add a sidebar to the "My Account" page. This sidebar should have a placeholder for the profile pic, a button to change the profile pic and a button to delete the account. These buttons don't need to be linked to any actions yet.

Note: Don't worry too much about the CSS and how it looks right now. We'll make it look nicer once we finalize the hi-fi designs in a future sprint.

<img width="154" alt="Screen Shot 2020-10-31 at 4 59 59 PM" src="https://user-images.githubusercontent.com/44531733/97792384-ad386580-1b9a-11eb-82a5-6ec4c39df98d.png">

Issue #25, opened 2020-10-31
Add "Reset Password" card on "My Account" page 


andrewzulaybar:
## 🔨 Task
Add the "Reset Password" section in a card on the "My Account" page and implement the following fields in a form. The form doesn't have to connect to anything yet, but we just want the skeleton set up.

Note: Don't worry too much about the CSS and how it looks right now. We'll make it look nicer once we finalize the hi-fi designs in a future sprint.

<img width="312" alt="Screen Shot 2020-10-31 at 4 57 12 PM" src="https://user-images.githubusercontent.com/44531733/97792358-49ae3800-1b9a-11eb-945c-e832b51ac53d.png">

Issue #24, opened 2020-10-31
Add "Account Details" card on "My Account" page


andrewzulaybar:
## 🔨 Task
Add the "Account Details" section in a card on the "My Account" page and implement the following fields in a form. The form doesn't have to connect to anything yet, but we just want the skeleton set up. 

Note: Don't worry too much about the CSS and how it looks right now. We'll make it look nicer once we finalize the hi-fi designs in a future sprint.

<img width="334" alt="Screen Shot 2020-10-31 at 4 52 29 PM" src="https://user-images.githubusercontent.com/44531733/97792293-a0ffd880-1b99-11eb-8e73-80342b630230.png">

PR #78, opened 2020-11-24
Setup MyAccount page

### LINKED ISSUE #23: Set up "My Account" route and screen
## 🔨 Task
Create an `/account` route and have that route navigate to a separate screen. We can just have simple links at the top of the page that navigate to this route for now until we get the top navigation bar implemented.

richardyjtian:
Capture

Finished the hi-fi frontend design, but it still needs to be integrated with backend.
@ericrkuo @yunishimura0716 do you have an idea on how the page would interact with the backend from the four buttons?

Resolves #23, #24, #25, #26

### DIFF
--- a/frontend/src/Router.tsx
+++ b/frontend/src/Router.tsx
@@ -5,6 +5,7 @@ import NavigationBar from "./components/navigation/NavigationBar";
 import { PromotionsListProvider } from "./contexts/PromotionsListContext";
 import Home from "./screens/Home";
 import MyPromotions from "./screens/MyPromotions"
+import MyAccount from "./screens/MyAccount"
 
 export default function Router(): ReactElement {
   return (
@@ -17,6 +18,9 @@ export default function Router(): ReactElement {
           <Route path="/login">
             {/* <Login /> */}
             <MyPromotions />
+          </Route>
+          <Route path="/myaccount">
+            <MyAccount />
           </Route>
           <Route path="/promotion/upload">{/* <UploadPromotion /> */}</Route>
           <Route path="/">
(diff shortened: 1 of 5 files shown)


ericrkuo:
Hey Richard, PR looks great, most of the stuff concernjng users is still WIP but here's what I know so far

For deleting account, that should just be a delete request with the user ID.

For updating the account details thatll be a patch request with user ID as well.

Both of those requests are still WIP in Yu's PR, but if u take a look at `UserRouter` that should contain how to format the requests.

For user photos, I have no idea ): we don't support photos at the moment. Id imagine that we can use AWS to store/upload/retrieve images and return a url back to the user with the image in the future.

For resetting password, that will be more difficult cause its still at its very early stages. @na-wu is working on the firebase implementation and that will most likely involve some new endpoints that we need.

richardyjtian:
> Hey Richard, PR looks great, most of the stuff concernjng users is still WIP but here's what I know so far
> 
> For deleting account, that should just be a delete request with the user ID.
> 
> For updating the account details thatll be a patch request with user ID as well.
> 
> Both of those requests are still WIP in Yu's PR, but if u take a look at `UserRouter` that should contain how to format the requests.
> 
> For user photos, I have no idea ): we don't support photos at the moment. Id imagine that we can use AWS to store/upload/retrieve images and return a url back to the user with the image in the future.
> 
> For resetting password, that will be more difficult cause its still at its very early stages. @na-wu is working on the firebase implementation and that will most likely involve some new endpoints that we need.

Thanks for your comments Eric! I'll definitely look into those.

andrewzulaybar:
Closing this PR because I transferred it over to https://github.com/ubclaunchpad/PromoPal-frontend/commit/b271bdd0432814eb23512856088c04261f232304

PR #81, opened 2020-11-28
Implemented Redis Caching for Restaurant Lat/Lon Values and Storing Restaurant Names in DB

ld357:
Purpose: To store the restaurant location (lat/lon) in a cache for 30 days and user-inputted restaurant name featured in a promotion in the DB. I also made small adjustments to the README and added parts for Redis :) 

Although not the best decision, we're storing the user-input for now because we need this info immediately in the FE and we don't want to issue multiple GooglePlacesAPI requests.

A FEW NOTES:
1. Currently, you need to change the 'host' variable inside createClient for the `CachingService.ts` file if you want to switch between Docker (use `redis-server`) or a local Redis server (use` local-host`). This isn't the most convenient so I've made an issue to address later. 
2. The ENV file with the GooglePlacesAPI key from FE was mistakenly exposed but I've disabled the key! 
3. More research needs to be done in terms of security and I've created an issue for it!

The decision to keep the GooglePlacesAPI on the FE was because it provides specific retrievals, a simpler implementation for the 30-day refresh as it doesn't require a GET request, and we eventually want client-specific caching.

This is basically what the flow will generally look like:

**FOR UPLOADING PROMOTIONS**
1. User submits restaurant name and location
2. Send Places API requests to get placeID and location details
3. Send POST request to save placeID, restaurant name and other details in DB and cache the lat/lon values 

**FOR HOMEPAGE**
1. User loads the home page
2. Send a GET /promotions request to get all the promotion information and retrieve all the cached location details 

**FOR SPECIFIC CLICKS ON RESTAURANT** 
1. Send a GET /promotion request to retrieve the promotion info and retrieve all the cached location details 
2. Send a Places API request with placeID to retrieve all other details

**THIRTY DAYS LATER**
_This still needs to be done_
1. Send a GET request to retrieve the placeID associated with the promotion
2. Send a Places API request to get the lat/lon values (needs to handle NOT_FOUND which is also another issue)
3. Send POST request to cache location details

I debated between choosing AWS ElastiCache with Redis vs. Docker and Redis for awhile and opted for the latter just because it was way simpler- the first option would require some work in setting up a VPN for local testing/development! Nathan also suggested that we can implement the rate limiter for Redis too!


backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,52 @@
+import redis from 'redis';
+import { CachingObject } from '../data/CachingObject';
+
+export class CachingService {
+  // todo: change to env later
+  private client = redis.createClient({
+    // host: 'redis-server',
+    host: 'localhost',
```
ld357, line 8: This is where you would change it between Docker and local Redis server!

backend/src/test/service/CachingService.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  // NOTE: RUNNING THESE TESTS WILL MODIFY THE REDIS SERVER!
+  let cachingService: CachingService;
+
+  beforeEach(() => {
+    cachingService = new CachingService();
+  });
+
+  afterEach(() => {
+    const client = cachingService.getClient();
+    client.quit();
```
ld357, line 14: I've left the Redis server open for the application since it takes resources to open- but open to more research and discussion as well 😄

ericrkuo:
> Send a Places API request with placeID to retrieve all other details
for this point, this is being done in FE right?

backend/src/main/controller/PromotionController.ts
```diff
@@ -39,14 +43,24 @@ export class PromotionController {
...
       ).getAllPromotions(promotionQuery);
+
+      for (const promotion of promotions) {
+        const currPlaceID = promotion.placeId;
+        const locationDetails: CachingObject = await this.cachingService.getLatLonValue(
+          currPlaceID
+        );
+        promotion.lat = locationDetails.lat;
+        promotion.lon = locationDetails.lon;
+      }
```
ericrkuo, line 54: What do you think about using `Promise.all`, mainly for this performance reason https://stackoverflow.com/a/46086037.

I would maybe even go one step further and pass in the promotion so that the `cachingService` would set the lat lon for the promotion, so then it could be something like

```suggestion
      const promises = promotions.map(promotion => this.cachingService.setLatLon(promotion));
      // then return all promotions with their lat lon values
```

And you can also see how much more is abstracted away since we don't really want to show too many implementation details inside the controller.

Let me know your thoughts on this!

backend/src/main/entity/Promotion.ts
```diff
@@ -144,4 +146,14 @@ export class Promotion {
...
+
+  /*
+   * Represents restaurant name inputted by user
+   * Many promotions can come from the same restaurant and thus have the same restaurantName
+   * */
+  @Column()
+  restaurantName: string;
+
+  lon?: number;
+  lat?: number;
```
ericrkuo, line 158: Maybe leave a comment why these ones are not columns?

backend/src/main/entity/Promotion.ts
```diff
@@ -144,4 +146,14 @@ export class Promotion {
     select: false,
   })
   tsVector: string;
+
+  /*
+   * Represents restaurant name inputted by user
+   * Many promotions can come from the same restaurant and thus have the same restaurantName
```
ericrkuo, line 152: Do you think it's okay to leave out this comment?

backend/src/main/service/CachingService.ts
ericrkuo: What do you think about creating `client` in App.ts? And then pass the client as a constructor parameter for Caching Service?

backend/src/main/validation/PromotionValidation.ts
```diff
@@ -21,6 +21,9 @@ export class PromotionValidation {
     name: Joi.string().required(),
     description: Joi.string().required(),
     expirationDate: Joi.date().required(), // note incoming format is a string, and Joi will automatically convert to type Date
+    lat: Joi.number().required(),
+    lon: Joi.number().required(),
```
ericrkuo, line 25: I remember you mentioned that if the input was a string, it would convert it to a number automatically. If that's expected behaviour, maybe you can put a comment above explaining that?

backend/src/test/repository/DiscountRepository.test.ts
```diff
@@ -24,7 +24,7 @@ describe('Unit tests for DiscountRepository', function () {
       fail('Should have failed');
     } catch (e) {
       expect(e.message).toBe(
-        'null value in column "promotionId" violates not-null constraint'
+        'null value in column "promotionId" of relation "discount" violates not-null constraint'
```
ericrkuo, line 27: Hmm strange, was this failing before?

backend/src/test/repository/EntityIntegration.test.ts
```diff
@@ -46,7 +46,7 @@ describe('Integration tests for all entities', function () {
       fail('Should have failed');
     } catch (e) {
       expect(e.message).toBe(
-        'null value in column "userId" violates not-null constraint'
+        'null value in column "userId" of relation "promotion" violates not-null constraint'
```
ericrkuo, line 49: Same comment here

backend/src/test/service/CachingService.test.ts
```diff
@@ -0,0 +1,41 @@
+import { CachingService } from '../../main/service/CachingService';
+import { CachingObject } from '../../main/data/CachingObject';
+
+describe('simple add/get for redis cache', function () {
+  // NOTE: RUNNING THESE TESTS WILL MODIFY THE REDIS SERVER!
```
ericrkuo, line 5: Hmm yeah this is a difficult one. Essentially, when we run tests, we don't want them to interfere with the cache because the cache may contain sensitive data that we don't want to mix with test data.

This would be a good opportunity to create a mock redis client, so for example, when we call `redis.setex` here, `redis` would actually be a mock instance and `setex` could just be a simple hashmap implementation for the mock class. Note, you'll need to have dependency injection in order to create a mock redis client. Let me know your thoughts, there's a lot of ways we can approach this!

https://github.com/ubclaunchpad/foodies/blob/423ff05514c8b0729ba4e3f4cadad9373d392f58/backend/src/main/service/CachingService.ts#L19-L30

backend/src/test/service/CachingService.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  test('getting lat/lon value', async () => {
+    return cachingService
+      .getLatLonValue('ChIJIfBAsjeuEmsRdgu9Pl1Ps48')
+      .then((result: CachingObject) => {
+        expect(result).toHaveProperty('lat');
+        expect(result).toHaveProperty('lon');
+        expect(result.lat).toBe(-34.2);
+        expect(result.lon).toBe(-46.123);
```
ericrkuo, line 35: I think the elvis operator could cover the `toHaveProperty` part. Do you think this will work?
```suggestion
        expect(result?.lat).toBe(-34.2);
        expect(result?.lon).toBe(-46.123);
```

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,52 @@
...
+          reject(err);
+        }
+
+        if (data) {
+          resolve(JSON.parse(data));
+        } else {
+          reject('No lat/lon information for this placeID');
+        }
+      });
+    });
```
ericrkuo, line 46: You mentioned that we don't cover the case yet of if the cache does not contain the placeId. Is this the place where we would add that behaviour?

LIke if the cache does not have the placeId, we would issue a request to get the new lat lon values so that we can cache again? Maybe we can leave a TODO or create an issue for that?

backend/src/main/validation/PromotionValidation.ts
ericrkuo, line 25: Also should we enforce something like this? https://stackoverflow.com/questions/15965166/what-is-the-maximum-length-of-latitude-and-longitude

ericrkuo (commented):
Excellent work looks really thorough and the code makes a lot of sense. Thanks for the detailed description!! Just some comments to get some discussion going on different design approaches we can take.

backend/ormconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "type": "postgres",
-  "host": "db",
+  "host": "localhost",
```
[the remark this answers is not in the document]
ld357: Gotcha!

backend/src/test/service/CachingService.test.ts
ld357: Hmm I'm not sure if that'll work since it's going to check if result is null/undefined but I'm checking that result has properties lat/lon?

backend/src/main/service/CachingService.ts
ld357: Hmm.. That's definitely do-able I think!! What were the benefits you were thinking of with regards to that?

ld357:
> > Send a Places API request with placeID to retrieve all other details
> 
> for this point, this is being done in FE right?

Yup!

backend/src/main/controller/PromotionController.ts
```diff
@@ -13,13 +13,17 @@ import {
   PromotionQueryDTO,
   PromotionQueryValidation,
 } from '../validation/PromotionQueryValidation';
+import { CachingService } from '../service/CachingService';
+import { CachingObject } from '../data/CachingObject';
 
 export class PromotionController {
+  private cachingService = new CachingService();
```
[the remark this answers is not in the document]
ld357: That could work! Where were you thinking of instantiating it then?

backend/src/main/controller/PromotionController.ts
ld357: Hmm yeah, it does look cleaner! I also didn't know that we shouldn't show too many details inside the Controller- why is that?

backend/src/test/service/CachingService.test.ts
ld357: Hmm that's a good point! I think one alternative I was also thinking of is just using a local Redis server for our tests? So we would switch in and out which client we're working with

backend/src/main/service/CachingService.ts
ld357: Yup! I was planning to create an issue for it but basically thinking of putting it in here and making a new request to the FE Google Places API service

backend/src/main/validation/PromotionValidation.ts
ld357: Hmm.. Since we're always getting it from Google Places API I think it's okay?

backend/src/test/repository/EntityIntegration.test.ts
ld357: Sorry 😅  which comment?

backend/src/main/service/CachingService.ts
ericrkuo: Thanks! Yeah so the main benefits I was thinking was like:

1. On application start-up, we can detect errors if our Redis client didn't connect rather than detecting errors with Redis when we instantiate and use `CachingService`
2. We maintain a single instance of the `client`. Because right now each time we instantiate `CachingService` we create a client. I'm not sure if `redis.createClient` takes a while or not.
3. I think it's also a bit easier to find in the future where we are connecting to our cache and database if they're at the same place

Let me know if this makes sense!

backend/src/main/controller/PromotionController.ts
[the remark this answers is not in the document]
ericrkuo: Right now we instantiate all our dependencies inside `App.ts`. In the future, I am thinking we can use Inversion of Control containers for all our service classes since the service classes need to be stateless.

Example: https://www.npmjs.com/package/inversify

backend/src/main/controller/PromotionController.ts
ericrkuo: >I also didn't know that we shouldn't show too many details inside the Controller- why is that?

Great question! I have very little experience, but I think cause generally, the controllers shouldn't really show too many implementation details. Right now we should show the minimum and delegate responsibility to our services. https://softwareengineering.stackexchange.com/questions/26438/how-much-business-logic-should-be-allowed-to-exist-in-the-controller-layer @andrewzulaybar maybe you can chime in on your thoughts for this too?

>I'm also unsure how to associate a particular promotion with their Caching Object with Promise.all?

Hmm I'm not sure what you mean.
I was thinking `CachingService` would then define it as
```
fun setLatLon(promotion: Promotion): Promise<Promotion> {
        const locationDetails: CachingObject = await this.cachingService.getLatLonValue(
          promotion.placeId
        );
        promotion.lat = locationDetails.lat;
        promotion.lon = locationDetails.lon;
        return promotion;
}
```

backend/src/test/repository/EntityIntegration.test.ts
ericrkuo: Oh I realize github hides comments when there's a lot, just above I think you need to click load more
image

backend/src/test/service/CachingService.test.ts
ericrkuo: Ah yup that would work just as great! Similar to our test database as well.

backend/src/test/service/CachingService.test.ts
ericrkuo: Ah wait sorry I mean if `result.lat` doesn't exist, then it will be undefined. So then 

```
expect(result.lat).tobe(-34.2)
expect(undefined).tobe(-34.2)
// throws error
```

But doesn't matter now that I think about it 😅

backend/src/main/service/CachingService.ts
ericrkuo: Ah yup sure I would recommend creating the issue and then linking the issue in a todo comment inside the code. For example,

```suggestion
    return new Promise((resolve, reject) => {
      this.client.get(placeId, (err: Error | null, data: string | null) => {
        if (err) {
          // todo: https://github.com/ubclaunchpad/foodies/issues/89 ... some more details
          reject(err);
        }

        if (data) {
          resolve(JSON.parse(data));
        } else {
          reject('No lat/lon information for this placeID');
        }
      });
    });
```

backend/src/main/service/CachingService.ts
ld357: Oooh yeah those are good points! For #2, I don't believe createClient takes awhile but given the benefits of #1/3 I can move it

backend/src/main/controller/PromotionController.ts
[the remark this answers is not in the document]
ld357: Gotcha! That makes sense, I'll check the link out and change it!

backend/src/main/controller/PromotionController.ts
ld357: Ooooh I see, thanks for the link!! And that makes sense, also wondering if that would still be considered Promise.all since we're using the JS map functionality?

backend/src/test/service/CachingService.test.ts
ld357: Awesome!! In that case, I'll go with that :)

backend/src/main/validation/PromotionValidation.ts
ericrkuo: I think in general we may not want to design our endpoints for how our FE uses it. But rather define how our FE should use it. Probably the decimal places might be too much, but in the future say we no longer use Google Places API lat/lon values and we generate our own. This may be something to validate, because I think if our lat lon was only 1 decimal place, then that's only accurate to 11 km 😝

backend/src/test/service/CachingService.test.ts
ld357: All good 😄 ! Does this look okay then?

backend/src/main/entity/Promotion.ts
ld357: Hmm yeah I can! I think it's nice to point out though?

backend/src/main/validation/PromotionValidation.ts
ld357: Yeah sure!!

backend/src/test/repository/DiscountRepository.test.ts
ld357: Yeah it was for some reason!

backend/src/main/validation/PromotionValidation.ts
ld357: LOOOL that's true!! I think we're def better off having more decimals than less, but yeah I think we can push it off for the future 😄

backend/src/main/service/CachingService.ts
ericrkuo: Sounds good thanks! Yah the reason why I thought 2 was because it takes a while for us to establish database connection for TYpeORM

backend/src/main/controller/PromotionController.ts
ericrkuo: Hmm like it would return an array of promises right and then all you would do is `await Promise.all(promsies)`. 

Now that I think about it, do you think we could abstract it even more and be like
```suggestion
      await this.cachingService.setLatLonForPromotions(promotions)
```

backend/src/main/entity/Promotion.ts
ericrkuo: Ah wait yes my bad sorry looks good! 😅

backend/src/test/repository/DiscountRepository.test.ts
ericrkuo: Hmm does this happen on `master` branch as well?

backend/src/test/service/CachingService.test.ts
ericrkuo: Yep looks great! Sorry for being so picky

backend/src/main/App.ts
```diff
@@ -47,10 +49,24 @@ createConnection()
...
+function handleRedisConnection(redisClient: RedisClient) {
+  redisClient.on('connect', function () {
+    console.log('Connected to Redis');
+  });
+  redisClient.on('error', function (err) {
+    console.log('Redis error: ' + err);
+  });
+}
+
 function registerRouters(app: Express) {
```
ericrkuo, line 61: think the `async` got removed when u resolved merge conflicts

yunishimura0716:
Well done!. All documents are readable and written very well. And also the high-level design for redis caching makes sense to me.

yunishimura0716 (approved):

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (err) {
+          // todo: https://github.com/ubclaunchpad/foodies/issues/89
+          reject(err);
+        }
+
+        if (data) {
+          return resolve(JSON.parse(data));
+        } else {
+          return resolve(JSON.parse(''));
+        }
```
andrewzulaybar, line 45: Very minor, but I think this can be simplified to:

```suggestion
        return resolve(JSON.parse(data ?? ''));
```

If you're expecting the second case to execute only when data is not `null` or `undefined`!

backend/src/test/controller/BaseController.ts
```diff
@@ -1,13 +1,21 @@
...
   const app = new App();
   const expressApp = express();
-  app.registerHandlersAndRoutes(expressApp);
+  await app.registerHandlersAndRoutes(expressApp, redisClient);
   return expressApp;
 };
+
+export const connectRedisClient = async (): Promise<RedisClient> => {
+  // see https://www.npmjs.com/package/redis-mock
+  return redisMock.createClient();
```
andrewzulaybar, line 20: Oooh nice that you found a mock for it :)

backend/src/test/validation/PromotionValidation.test.ts
```diff
@@ -383,13 +386,16 @@ describe('Unit tests for PromotionValidation', function () {
         name: 3,
         placeId: 4,
         userId: false,
+        lat: '34.0',
+        lon: '-43.2',
```
andrewzulaybar, line 390: Should these be numbers? Or should lines 44-45 be strings? Or does it not matter 😅

Wondering if this is related to the comment you made in `backend/src/main/validation/PromotionValidation.ts`

andrewzulaybar (approved):
This looks great!!! Thanks for your hard work on this, looking forward to make use of it soon :)

I made a few comments but besides that, there just seem to be a few lint warnings if you want to quickly handle those!

backend/src/main/App.ts
```diff
@@ -51,12 +54,17 @@ export class App {
...
+  async registerHandlersAndRoutes(
+    app: Express,
+    redisClient: RedisClient
+  ): Promise<void> {
     app.use(bodyParser.json());
 
     app.get('/', (req, res) => res.send('Hello World'));
 
-    const promotionController = new PromotionController();
+    const cachingService = new CachingService(redisClient);
```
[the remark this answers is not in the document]
ld357: Props to @ericrkuo for suggesting it!

backend/src/main/resources/Data.ts
```diff
@@ -114,7 +114,8 @@ const promotion1 = new Promotion(
   It’s never been easier to give back - all you have to do is eat your favourite fries (120-560 cals)! You can order them in-restaurant, at the drive thru, through McDelivery or you can order ahead on the McDonald’s app.
   If you’re looking for more ways to help families with sick children you can always round up your order - even when Fries for Good is over. When you finish placing your order, simply ask to “Round Up for RMHC” at participating McDonald’s restaurants, and your order will be rounded to the nearest dollar. The difference will be donated to RMHC Canada.`,
   new Date(),
-  new Date()
+  new Date(),
+  'promo1'
```
ericrkuo, line 118: Hehe sorry do you mind filling these in with restaurant names,  cause right now the sample data tries its best to use real descriptions/names

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Promotion } from '../entity/Promotion';
+
+export class CachingService {
+  private client: RedisClient;
+
+  constructor(client: RedisClient) {
+    this.client = client;
+  }
+
+  async cacheLatLonValues(
```
ericrkuo, line 12: Could we add documentation for each of the functions here? And also one that summarizes the class?

backend/src/test/controller/BaseController.ts
ld357: Eric actually found this one! It turned out to be really simple which is nice :)

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,71 @@
...
+    placeId: string,
+    latitude: number,
+    longitude: number
+  ): Promise<boolean> {
+    return new Promise((resolve, reject) => {
+      const value: CachingObject = { lat: latitude, lon: longitude };
+      this.client.setex(
+        placeId,
+        2592000,
```
ericrkuo, line 21: Maybe a comment what `2592000` represents?

backend/src/test/validation/PromotionValidation.test.ts
ld357: It is! It'll actually be converted to numbers ☺️

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,71 @@
...
+          }
+          return resolve(true);
+        }
+      );
+    });
+  }
+
+  async getLatLonValue(placeId: string): Promise<CachingObject> {
+    return new Promise((resolve, reject) => {
+      this.client.get(placeId, (err: Error | null, data: string | null) => {
```
ericrkuo, line 35: Just curious, how come the parameters are `(err: Error | null, data: string | null)`

backend/src/test/controller/EnumController.test.ts
```diff
@@ -1,21 +1,25 @@
...
     await connection.create();
-    app = registerTestApplication();
+    redisClient = await connectRedisClient();
+    app = await registerTestApplication(redisClient);
   });
 
   afterAll(async () => {
     await connection.close();
+    redisClient.quit();
   });
```
ericrkuo, line 24: Yeah now the only con I think with this is that we need to use `redis` even for controller tests that don't involve redis. What would happen if we made `redisClient` an optional parameter to `registerTestApplication` and left the class the way it was before?

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -40,9 +44,16 @@ describe('Unit tests for PromotionController', function () {
     const promotion = new PromotionFactory().generate(user, discount, [
       schedule,
     ]);
```
ericrkuo, line 46: Could we modify the factory method to set the `lat/lon` values for us? Because I noticed that a lot of the tests now have to maintain the responsibility to remember setting lat lon values for a promotion. What do you think about the suggestion?

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (err) {
+          // todo: https://github.com/ubclaunchpad/foodies/issues/89
+          reject(err);
+        }
+
+        if (data) {
+          return resolve(JSON.parse(data));
```
ericrkuo: Oh strange, the `data` returned needs to call `JSON.parse`?

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,71 @@
...
+      this.setLatLonForPromotion(promotion)
+    );
+    return Promise.all(promises);
+  }
+
+  async setLatLonForPromotion(promotion: Promotion) {
+    return this.getLatLonValue(promotion.placeId)
+      .then((locationDetails: CachingObject) => {
+        promotion.lat = locationDetails.lat;
+        promotion.lon = locationDetails.lon;
```
ericrkuo: Wouldn't this throw an error if `getLatLonValue` returned `JSON.parse('')`?

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -119,6 +146,29 @@ describe('Unit tests for PromotionController', function () {
...
+    await redisClient.setex(
+      promotion2.placeId,
+      2592000,
+      JSON.stringify({ lat: promotion2.lat, lon: promotion2.lon })
+    );
+    await redisClient.setex(
+      promotion3.placeId,
+      2592000,
+      JSON.stringify({ lat: promotion3.lat, lon: promotion3.lon })
+    );
```
ericrkuo, line 170: Do you think there is some way we can minimize the duplicated logic?

backend/src/main/service/CachingService.ts
ld357: Ooo, thank you! :))

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -182,6 +242,34 @@ describe('Unit tests for PromotionController', function () {
...
+          .get(`/promotions/${expectedPromotion.id}`)
+          .expect(200)
+          .end((err, res) => {
+            if (err) return done(err);
+            const promotion = res.body;
+            comparePromotions(promotion, expectedPromotion);
+            done();
+          });
+      });
+  });
```
ericrkuo, line 271: What does this test verify?

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -238,6 +330,29 @@ describe('Unit tests for PromotionController', function () {
...
+      user,
+      new DiscountFactory().generate(DiscountType.PERCENTAGE),
+      [new ScheduleFactory().generate()]
+    );
+
+    await userRepository.save(user);
+    request(app)
+      .post('/promotions')
+      .send({ ...expectedPromotion, user: undefined, userId: user.id })
+      .expect(201)
```
ericrkuo, line 345: How come it's a `201` when there is an error?

backend/src/test/service/CachingService.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  beforeEach(() => {
+    const redisMockClient = redisMock.createClient();
+    cachingService = new CachingService(redisMockClient);
+  });
+
+  afterEach(() => {
+    const client = cachingService.getClient();
+    client.quit();
+  });
```
ericrkuo, line 23: What do you think about `beforeAll` and `afterAll` instead?

backend/src/test/service/CachingService.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      .then((result: CachingObject) => {
+        expect(result).toHaveProperty('lat');
+        expect(result).toHaveProperty('lon');
+        expect(result.lat).toBe(-34.2);
+        expect(result.lon).toBe(-46.123);
+      })
+      .catch((e) => {
+        fail('Did not expect to fail: ' + e.message);
+      });
+  });
```
ericrkuo, line 48: Oh does this test depend on the test before?

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -5,26 +5,30 @@ import connection from '../repository/BaseRepositoryTest';
...
 import { PromotionRepository } from '../../main/repository/PromotionRepository';
 import { DiscountType } from '../../main/data/DiscountType';
 import { Promotion } from '../../main/entity/Promotion';
+import { RedisClient } from 'redis-mock';
 
 describe('Unit tests for PromotionController', function () {
   let userRepository: UserRepository;
   let promotionRepository: PromotionRepository;
   let app: Express;
+  let redisClient: RedisClient;
```
ericrkuo, line 21: Could we call this `mockRedisClient`?

backend/src/main/data/CachingObject.ts
```diff
@@ -0,0 +1,4 @@
+export interface CachingObject {
+  lat: number;
+  lon: number;
+}
```
ericrkuo, line 4: Maybe some comments as well describing the purpose of the interface?

ericrkuo (commented):
Really great work @ld357!!! Sorry, there may seem like a lot of review comments, but these are mainly just for further discussion and some to help with my understanding! I really like how well the mock Redis works and thanks for all the great work for CI, docker-compose and the tests! Really great quality code!

backend/src/test/validation/PromotionValidation.test.ts
ericrkuo: Oh how come we can't just enforce that the lat and lon will be numbers and cannot be strings?

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -182,6 +242,34 @@ describe('Unit tests for PromotionController', function () {
...
+      .send({ ...expectedPromotion, user: undefined, userId: user.id })
+      .then(async () => {
+        await promotionRepository.save(expectedPromotion);
+        request(app)
+          .get(`/promotions/${expectedPromotion.id}`)
+          .expect(200)
+          .end((err, res) => {
+            if (err) return done(err);
+            const promotion = res.body;
+            comparePromotions(promotion, expectedPromotion);
```
ericrkuo, line 267: Ohya and I think we should update `comparePromotions` to check that the lat/lon are being returned

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,85 @@
...
+    const placeId: string = promotion.placeId;
+    const latitude: number = promotion.lat;
+    const longitude: number = promotion.lon;
+    const result = await this.cacheLatLonValues(placeId, latitude, longitude);
+    return result;
+  }
+
+  // caches the lat/lon values for a promotion's restaurant
+  // key is placeId and value is of the form { lat: x, lon: y }
+  // sets the cached value to expire in 30 days (limit imposed by Google Places API)
```
ericrkuo, line 26: Minor comment, could we follow the format for the way comments are structured for `PromotionController` and our other classes?

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  // formats promotionDTO for caching
+  async getValuesAndCache(promotion: PromotionDTO): Promise<boolean> {
+    const placeId: string = promotion.placeId;
+    const latitude: number = promotion.lat;
+    const longitude: number = promotion.lon;
+    const result = await this.cacheLatLonValues(placeId, latitude, longitude);
+    return result;
+  }
```
ericrkuo, line 22: I don't think we'll need this method for my suggestion in `PromotionController::addPromotion`

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,85 @@
...
+        }
+        return resolve(JSON.parse(data ?? '{}'));
+      });
+    });
+  }
+
+  // returns redisClient
+  getClient(): RedisClient {
+    return this.client;
+  }
```
ericrkuo, line 64: I think this is only used in our tests. I think if you structure the `beforeAll` and `afterAll` like you did in`PromotionController` you dont need this method

backend/src/test/service/CachingService.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      .then(() => {
+        const expectedPromotion = promotion;
+        expectedPromotion.lat = -34.2;
+        expectedPromotion.lon = -46.123;
+        expect(expectedPromotion).toEqual(promotion);
+      })
+      .catch((e) => {
+        fail('Did not expect to fail: ' + e.message);
+      });
+  });
```
ericrkuo, line 105: You don't need the `async` keyword if `await` is not used. Do you think we could use `await` with try catches just like how our other tests are formatted

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,85 @@
...
+        JSON.stringify(value),
+        (err: Error | null) => {
+          if (err) {
+            return reject(err);
+          }
+          return resolve(true);
+        }
+      );
+    });
+  }
```
ericrkuo, line 46: How come we are returning a `boolean`? It seems like this value is only used for tests and we never return `resolve(false)` or `reject(false)`. We could return `Promise<void>` since if it's successful we know it didn't fail with a promise reject.

backend/src/main/service/CachingService.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  // sets the lat/lon value for a promotion
+  async setLatLonForPromotion(promotion: Promotion): Promise<void> {
+    return this.getLatLonValue(promotion.placeId)
+      .then((locationDetails: CachingObject) => {
+        promotion.lat = locationDetails?.lat;
+        promotion.lon = locationDetails?.lon;
+      })
+      .catch((err) => {
+        throw err;
```
ericrkuo, line 82: What happens here when the `err` is thrown and what does the format of the error message look like when we call the endpoint?

backend/src/main/validation/PromotionValidation.ts
```diff
@@ -44,6 +44,9 @@ export class PromotionValidation {
       .default(new Date()),
     /** note incoming format is a string, and Joi will automatically convert to type Date */
     expirationDate: Joi.date().required(),
+    lat: Joi.number().required().strict(true), // note incoming format is a string, and Joi will automatically convert to type Number (both lat/lon)
```
ericrkuo, line 47: `// note incoming format is a string, and Joi will automatically convert to type Number (both lat/lon)`
I think this comment is not needed anymore

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -74,6 +83,17 @@ describe('Unit tests for PromotionController', function () {
...
+    await mockRedisClient.setex(
+      promotion1.placeId,
+      2592000,
+      JSON.stringify({ lat: promotion1.lat, lon: promotion1.lon })
+    );
+    await mockRedisClient.setex(
+      promotion2.placeId,
+      2592000,
+      JSON.stringify({ lat: promotion2.lat, lon: promotion2.lon })
+    );
```
ericrkuo, line 95: I think there is still a lot of duplicated code for this test class.

I was thinking something like this
```
await setLatLon([promotion1, promotion2])
```

and you define the function `setLatLon`. What do you think?

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -203,10 +242,12 @@ describe('Unit tests for PromotionController', function () {
       .end((err, res) => {
         const frontEndErrorObject = res.body;
         expect(frontEndErrorObject?.errorCode).toEqual('ValidationError');
-        expect(frontEndErrorObject.message).toHaveLength(1);
+        expect(frontEndErrorObject.message).toHaveLength(3);
         expect(frontEndErrorObject.message[0]).toContain(
           '"cuisine" must be one of'
         );
+        expect(frontEndErrorObject.message[1]).toContain('"lat" is required');
+        expect(frontEndErrorObject.message[2]).toContain('"lon" is required');
```
ericrkuo, line 250: I don't think this is needed since you created a test `'POST /promotions/ - should not be able to add promotion if lat/lon do not exist'` that seems to test the same thing

backend/src/test/service/CachingService.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      })
+      .then(() => {
+        const expectedPromotion1 = promotion1;
+        expectedPromotion1.lat = -34.2;
+        expectedPromotion1.lon = -46.123;
+        expect(expectedPromotion1).toEqual(promotions[0]);
+        const expectedPromotion2 = promotion2;
+        expectedPromotion2.lat = -34.2;
+        expectedPromotion2.lon = -46.123;
+        expect(expectedPromotion2).toEqual(promotions[1]);
```
ericrkuo, line 75:
```suggestion
   for (let promotion of promotions) {
        expect(promotion .lat).toEqual(TEST_LAT);
        expect(promotion .lon).toEqual(TEST_LON);
   }
```

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -350,6 +417,8 @@ describe('Unit tests for PromotionController', function () {
       placeId: expectedPromotion.placeId,
       expirationDate: expectedPromotion.expirationDate.toISOString(),
       startDate: expectedPromotion.startDate.toISOString(),
+      lat: expectedPromotion.lat,
+      lon: expectedPromotion.lon,
```
ericrkuo, line 421: Nice!

backend/src/test/service/CachingService.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  afterAll(() => {
+    const client = cachingService.getClient();
+    client.quit();
+  });
+
+  test('caching and getting lat/lon values', () => {
+    return cachingService
+      .cacheLatLonValues('ChIJIfBAsjeuEmsRdgu9Pl1Ps48', -34.2, -46.123)
```
ericrkuo, line 27: What do you think about using constants throughought this test class
```suggestion
      .cacheLatLonValues(TEST_PLACE_ID, TEST_LAT, TEST_LON)
```

ericrkuo (commented):
Great job @ld357! Just some quick clean up and some minor comments. Let me know if you need clarification on any of the comments 😄

backend/src/main/App.ts
```diff
@@ -54,8 +58,11 @@ export class App {
       });
       const app = express();
 
-      // await this.loadSampleData();
-      this.registerHandlersAndRoutes(app);
+      this.redisClient = await this.createRedisClient();
+      await this.registerHandlersAndRoutes(app, this.redisClient);
+
+      // load sample data and cache the lat/lon for existing data
+      await this.loadAndCacheSampleData();
```
ericrkuo, line 65: Could we keep this commented out?

backend/src/main/App.ts
```diff
@@ -162,6 +173,30 @@ export class App {
...
+      await cachingService.cacheLatLonValues(
+        promotion.placeId,
+        promotion.lat,
+        promotion.lon
+      );
+    }
+  }
+
+  async loadAndCacheSampleData(): Promise<void> {
+    // await this.loadSampleDBData();
```
ericrkuo, line 199: Can we keep both uncommented out?

ericrkuo (approved):

Issue #65, opened 2020-11-17
TypeORM add transactions


ericrkuo:
Currently, we do not have any transactions in place. For example, when we add a promotion, we need to communicate with the database to get the user, and then save the promotion and discount. This is very fragile for data integrity (say the user was deleted from another query **after** we queried for the user and **before** we are saving the promotion and discount). this will also make sure we maintain ACID, especially atomicity and isolation levels because we want to ensure certain operations are safe from dirty reads, non-repeatable reads, or phantom tuples, 

Basically, we need to ensure that when we query the database multiple times, they should all be under one transaction and we may need to set the isolation levels accordingly

TypeORM can support transactions, although we will need to restructure all our controllers. https://typeorm.io/#/transactions

@yunishimura0716  @andrewzulaybar what do you guys think about this?
This issue can be handled after Yu and I merge in our PRs.

Whoever takes the ticket will be responsible for documenting what transactions have been added!

andrewzulaybar:
@ericrkuo Yeah, this sounds like a good idea! Do you think we need to define transactions for all DB interactions that we have right now? Either way, I'm thinking we should create some kind of documentation (maybe it just needs to be comments) for all the latency-sensitive actions that can be performed on our data currently

ericrkuo:
I think we may need to for each one, but I'm not too sure.

I'm not sure if typeorm can mark some transactions as read-only, that would speed up performance.

But for all our transactions I'm thinking of setting them to the strictest isolation level (maybe it is the default for postres already)

And definitely yeah we can create some documentation, I'll update the ticket to encorproate that

ericrkuo:
Should be really simple to add transactions: 
https://typeorm.io/#/transactions
https://typeorm.io/#/custom-repository/using-custom-repositories-in-transactions-or-why-custom-repositories-cannot-be-services

just need to wrap all of our controller methods with transactions

yunishimura0716:
I'm just curious what happen for typeORM if we want to save promotion and discount but no user? It should throw the error, I think.

ericrkuo:
Ohyea it will throw error since it violates foreign key constraint. But ideally, we should detect this problem beforehand so that we can throw a more user friendly message

yunishimura0716:
Yeah, that's true. so, maybe we should change all controller which works with more than one table.

ericrkuo:
[...]

PR #82, opened 2020-11-28
65 - TypeORM add transactions

ericrkuo:
https://github.com/ubclaunchpad/foodies/issues/65

Changed all points of our application that communicate with our database to be inside transactions.

This is to ensure https://www.ibm.com/support/knowledgecenter/SSGMCP_5.4.0/product-overview/acid.html.

For example, if we have one method that first saves a user, and then saves a promotion. If we save the user successfully but fail to save the promotion, then the entire transaction is rolled back (nothing is persisted).


ld357 (approved):
Looks great! Thanks for sharing the docs :)

yunishimura0716:
Thank for your great work, looks useful and consistent design on controllers makes sense to me.

yunishimura0716 (approved):

Issue #64, opened 2020-11-17
Postgres Full Text Search


ericrkuo:
Right now we have a barebones implementation of Postgres full-text search. Essentially, this will enhance our search query functionality, when we search "Sale Chinese Dumplings", Postgres should be able to search all promotion descriptions and names in milliseconds. It would also be able to rank each one to relevance. 

See this comment for more details: https://github.com/ubclaunchpad/foodies/pull/52#pullrequestreview-530941149

What we have right now
- tsvector column
- migration to add trigger function, and gin index
- endpoint for user to submit search query and get promotions along with their rank

What we need to do
- tweak the search query functionality
- make sure we handle cases for escape characters (%, $, etc.)
- can place more preference (weights) on name than a description
- we can even highlight results
- OR AND filters
- modify the name and description of each Promotion to include more data (not just promoX or descriptionX). That way can test out whether search query works

Issue #54, opened 2020-11-12
Postgres or TypeORM is not storing the dates correctly.


ericrkuo:
![image](https://user-images.githubusercontent.com/49849754/98904510-9b927000-246e-11eb-8de9-b3fe9b6dcb4a.png)
The time of these are not correct, I'm pretty sure I didn't insert them at 3 am. I remember reading about this either in typeorm or for postgres. 

Not sure if anyone is seeing the same problem. I think we can fix it, I didn't research it yet.

This will block us from doing anything with dates, for example if we want to filter promotions by expiration dates, then we need the dates in our BE to be stored correclty. Hoping it should be a quick fix

ericrkuo:
This is a Postgres feature, **not a bug:** https://stackoverflow.com/questions/51132868/incorrect-date-fetched-from-postgres

Postgres always stores in UTC format (check link). E.g. if right now is 6:01 pm at our timezone, postgres stores it as 2020-11-23 02:01:00 as UTC format

Article: https://medium.com/building-the-system/how-to-store-dates-and-times-in-postgresql-269bda8d6403
Essentially, we should be able to still keep it in UTC format and get it back as UTC format. I think the idea is that ALL times stored in the DB will be converted to UTC, that way if our app is used somewhere across the world, the time can be converted to user's respective time zone.

https://spin.atomicobject.com/2015/12/09/time-zones-software-development/
This article recommends we use a library possible in our front end to convert these dates to the local time zone of the user

We could do something like this in FE to show dates correctly in the clients time zone:
https://stackoverflow.com/questions/38900413/localize-utc-time-date
![image](https://user-images.githubusercontent.com/49849754/99925602-a57c6480-2cf3-11eb-8e71-b21c6fd64de5.png)

In the BE if we want to compare dates, we can do something like this:
![image](https://user-images.githubusercontent.com/49849754/99926542-eaee6100-2cf6-11eb-91bc-4cb5b1f54cf1.png)

That way if all our data is UTC (photo below), we can do date comparisons as long as the date has its timezone,
![image](https://user-images.githubusercontent.com/49849754/99926558-fc376d80-2cf6-11eb-802d-18d692588475.png)
We can query with different time zones and these would all work since they translate to `2020-11-23 02:02:00` in UTC
```
2020-11-22 21:02:00 EST
2020-11-22 18:02:00 PST
```

ericrkuo:
Another alternative to `moment` for FE, could be something like this i think
[...]

PR #84, opened 2020-11-28
64, 54 - Full Text Search and Fix Query By Date

ericrkuo:
https://github.com/ubclaunchpad/foodies/issues/64, https://github.com/ubclaunchpad/foodies/issues/54

When we call GET `promotions?searchQuery=beef $30`, we can now query using Postgres's full-text search to simulate a low-level tier google search engine.

This means that results are ordered by relevance to the search query, and we even have parts of the promotion name and description that are bolded when keywords match the search query. Postgres's full-text search would also find words based on lexemes, so for example, if our search query is `matches`, Postgres can find other words like `match, matching`.

Example response here. Then in FE, we can show the bold description and bold name.
image

Also fixed querying by expiration date. We now first convert all times to UTC and then compare them. For some reason, although `promotion.expirationDate` is already in UTC format, I still needed to include `at time zone UTC`. 

We can now filter promotions by expiration date accurately up to even the `seconds`.
https://github.com/ubclaunchpad/foodies/blob/c37de590a6144c4241fb5f80c9bb86c22315c842/backend/src/main/repository/PromotionRepository.ts#L55

Also updated all our sample data!


yunishimura0716:
Great work! I'm just asking you about the interface of FullTextSearch. How do you get the value of `rank`? I'm sorry If I'm missing something

ericrkuo:
> Great work! I'm just asking you about the interface of FullTextSearch. How do you get the value of `rank`? I'm sorry If I'm missing something

Thanks! No need to apologize, the `rank` comes from here.
https://github.com/ubclaunchpad/foodies/blob/c37de590a6144c4241fb5f80c9bb86c22315c842/backend/src/main/repository/PromotionRepository.ts#L79-L83

So we use `ts_rank_cd` which takes in two inputs, a `ts_vector` and `ts_query`. We optimized FullTextSearch by adding a new column to promotions called `tsvector` which we use as the first input to `ts_rank_cd`. And our `ts_query` comes from the user's search query they input. Then we leave the rest to Postgres to compute the rank!

yunishimura0716:
> > Great work! I'm just asking you about the interface of FullTextSearch. How do you get the value of `rank`? I'm sorry If I'm missing something
> 
> Thanks! No need to apologize, the `rank` comes from here.
> https://github.com/ubclaunchpad/foodies/blob/c37de590a6144c4241fb5f80c9bb86c22315c842/backend/src/main/repository/PromotionRepository.ts#L79-L83
> 
> So we use `ts_rank_cd` which takes in two inputs, a `ts_vector` and `ts_query`. We optimized FullTextSearch by adding a new column to promotions called `tsvector` which we use as the first input to `ts_rank_cd`. And our `ts_query` comes from the user's search query they input. Then we leave the rest to Postgres to compute the rank!

I see, thanks. what does ts_vector represent?

ericrkuo:
ts_vector is basically a document in a form thats optimized for full text search. So like it's broken down into something called lexemes.
https://www.postgresql.org/docs/10/datatype-textsearch.html This link does a really good explanation.

I configured in the BE that ts_vector contains lexemes for a promotions **name and description**, so the ts_vector represents two columns. It's really powerful, that's how postgres knows that words like `fry, fries, frying, fried` are all very similar.

https://github.com/ubclaunchpad/foodies/blob/c37de590a6144c4241fb5f80c9bb86c22315c842/backend/src/main/migration/1605475222018-FullTextSearch.ts#L15

yunishimura0716:
> ts_vector is basically a document in a form thats optimized for full text search. So like it's broken down into something called lexemes.
> https://www.postgresql.org/docs/10/datatype-textsearch.html This link does a really good explanation.
> 
> I configured in the BE that ts_vector contains lexemes for a promotions **name and description**, so the ts_vector represents two columns. It's really powerful, that's how postgres knows that words like `fry, fries, frying, fried` are all very similar.
> 
> https://github.com/ubclaunchpad/foodies/blob/c37de590a6144c4241fb5f80c9bb86c22315c842/backend/src/main/migration/1605475222018-FullTextSearch.ts#L15

That's so cool! It's kind of technique of natural language processing! thank you so much for explaining!

yunishimura0716 (approved):

ericrkuo:
Oh very interesting, I did not see that connection! I don't know too much about natural language processing, but I can see how ts_vector could represent some kind of pre-processing and cleaning of the data. Do you mind explaining or clarifying how this connects to NLP techniques based on your experience?

yunishimura0716:
> Oh very interesting, I did not see that connection! I don't know too much about natural language processing, but I can see how ts_vector could represent some kind of pre-processing and cleaning of the data. Do you mind explaining or clarifying how this connects to NLP techniques based on your experience?

I'm not also expert of NLP and I just have used some library of NLP before. The thing ts_vector contains lexemes is similar to Morphological Analysis of NLP. Morphological Analysis is a kind of analysis in NLP, which makes Morpheme from text data based on the grammar and part of speech of the words. This analysis has been applied to many field of NLP like calculating the similarity between two sentences. I have used `Mecab`. If you're interested in, check it!

ericrkuo:
Woahh i see very interesting thanks for the explanation. I'll check it out!

backend/src/main/repository/PromotionRepository.ts
```diff
@@ -50,10 +50,14 @@ export class PromotionRepository extends Repository<Promotion> {
       });
     }
 
+    // see https://github.com/ubclaunchpad/foodies/issues/54
     if (promotionQuery?.expirationDate) {
-      queryBuilder.andWhere('promotion.expirationDate >= :date', {
-        date: promotionQuery.expirationDate,
-      });
+      queryBuilder.andWhere(
+        "promotion.expirationDate ::timestamptz at time zone 'UTC' >= :date ::timestamptz at time zone 'UTC'",
```
ld357, line 56: Sorry if you explained this in the past and I'm forgetting but what's the purpose of this?

ld357 (commented):
Nicely done!! This is so cool :) just had one question!

backend/src/main/repository/PromotionRepository.ts
ericrkuo: Oh this is if we want to filter promotions by expiration date, so for example if a user wants to find all promotions that are not expired after a given date. Does that answer your question?

ld357: Oooh I see, yup that does!

ld357 (approved):

Issue #49, opened 2020-11-07
Spike: research how we are going to implement upvote/downvote functionality


ericrkuo:
Edit** Most likely not needed for MVP yet, depending on time.

This will require some database changes in the BE and also need to talk with design team how we are going to design this feature.

Maybe frontend can detect whether user is spamming it, and once the user has left the button for X amount of seconds, then we send a request to BE to update the votes.

Research:
- TypeORM has repository increment/decrement functionality

na-wu:
Since we're using express, we could set up a rate limiter as middleware to reduce any sort of backend overload. I wrote a similar one here that could probably be modified for our use case

https://github.com/na-wu/qrapi/blob/master/ratelimiter.js

ericrkuo:
Totally agree, I was thinking along the same lines, that would be great. But how about cases if a user likes 5 separate promotions very fast. Do we want to batch these operations altogether and make it one transaction to save all 5 promotions at once?

na-wu:
I think handling 5 operations in a rapid manner should be pretty harmless, the way I imagine it is a user shouldn't notice any rate limiting under normal use cases, nor should we prevent users from saving multiple promotions in a short period of time. The number of operations allowed per second can probably be adjusted to what we might consider "normal".

ericrkuo:
That makes sense yah

PR #86, opened 2020-11-30
49 - Basic voting functionality

ericrkuo:
https://github.com/ubclaunchpad/foodies/issues/49

New column `votes` added to `Promotion`

Two new endpoints
```
POST promotions/:id/upVote
   // SQL statement: Update promotion set votes = votes + 1 where id = :id
POST promotions/:id/downVote
   // SQL statement: Update promotion set votes = votes - 1 where id = :id
```

I did not implement any rate-limiting yet, I think this is something we can tackle separately once we see a higher number of users for our product.


ericrkuo:
self reminder: add transactions after https://github.com/ubclaunchpad/foodies/pull/82 is merged

ld357 (approved):
Looks good! Are we planning to implement the getVotesPromotion in another PR?

ericrkuo:
> Looks good! Are we planning to implement the getVotesPromotion in another PR?

Do you mean when we call `GET /promotions or GET /promotions/:id` we return the `votes` as well? Or a separate endpoint that just returns the field `votes` for a promotion.

backend/src/main/controller/PromotionController.ts
```diff
@@ -164,4 +164,50 @@ export class PromotionController {
...
+    next: NextFunction
+  ): Promise<any> => {
+    try {
+      await getManager().transaction(async (transactionalEntityManager) => {
+        const id = await IdValidation.schema.validateAsync(request.params.id, {
+          abortEarly: false,
+        });
+        await transactionalEntityManager
+          .getCustomRepository(PromotionRepository)
+          .increment({ id: id }, 'votes', 1);
```
andrewzulaybar, line 183:
```suggestion
          .increment({ id }, 'votes', 1);
```

Super minor improvement here but also not a big deal!

backend/src/main/controller/PromotionController.ts
```diff
@@ -164,4 +164,50 @@ export class PromotionController {
...
+    next: NextFunction
+  ): Promise<any> => {
+    try {
+      await getManager().transaction(async (transactionalEntityManager) => {
+        const id = await IdValidation.schema.validateAsync(request.params.id, {
+          abortEarly: false,
+        });
+        await transactionalEntityManager
+          .getCustomRepository(PromotionRepository)
+          .decrement({ id: id }, 'votes', 1);
```
andrewzulaybar, line 206:
```suggestion
          .decrement({ id }, 'votes', 1);
```
Same as above

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -273,6 +273,72 @@ describe('Unit tests for PromotionController', function () {
...
+
+    request(app)
+      .post(`/promotions/${promotion.id}/downVote`)
+      .expect(204)
+      .end((err) => {
+        if (err) return done(err);
+        return getManager().transaction(
+          'READ UNCOMMITTED',
+          async (transactionalEntityManager) => {
+            // check that promotion votes has incremented
```
andrewzulaybar, line 328:
```suggestion
            // check that promotion votes has decremented
```
Another minor change!

andrewzulaybar (approved):
Just some minor suggestions and I also had a question: this implementation doesn't keep track of which user liked the promotion right? I'm thinking that this would be handled in a future improvement once we have login functionality and a way to determine which user is sending requests + prevent a user from liking it more than once.

Anyways, looks good to me and feel free to merge!

backend/src/test/controller/PromotionController.test.ts
ericrkuo: Oopsies thanks for catching this!

ericrkuo:
> Just some minor suggestions and I also had a question: this implementation doesn't keep track of which user liked the promotion right? I'm thinking that this would be handled in a future improvement once we have login functionality and a way to determine which user is sending requests + prevent a user from liking it more than once.
> 
> Anyways, looks good to me and feel free to merge!

Yeah actually this is something we should discuss. Currently, we have the behaviour to save a promotion, so we have a mapping of which user saved which promotions. I'm thinking that can be separate from the concept of `votes` which is more so like a gauge for credibility (similar to Reddit posts).

image
So for example, the upvote/downvote of a Reddit post can be translated to the credibility of a promotion with votes. And saving a post would be equivalent to saving a promotion. 

Hmm, but now that I think about it, you're right in terms of how we don't prevent a user from voting more than once because we don't know whether a user upvoted/downvoted a promotion already... Maybe I can create a table that has a `userid` and `promotionId`, and it tells which user upvoted or downvoted a promotion

PR #90, opened 2020-12-23
Setup CI

ericrkuo:
The first phase of adding continuous integration to our project using Github Actions
https://docs.github.com/en/free-pro-team@latest/actions

What is CI?
https://www.atlassian.com/continuous-delivery/continuous-integration

How does it work?
- Essentially we create a YAML file inside `.github/workflows`
- This YAML file is a workflow, which defines a set of jobs, each job consists of multiple steps
  - https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions
- We have three jobs defined
  - `lint` --> checks for linting errors
  - `build` --> ensures project builds correctly
  - `integration_tests` --> runs our integration tests
- how the integration tests work
  - We need some way to run a Postgres container since our integration tests depend on the database `foodies_test`
  - This can be solved with https://docs.github.com/en/free-pro-team@latest/actions/guides/creating-postgresql-service-containers
  - We define our service as Postgres and use the Postgres Docker image. We map the ports between the service container and the Docker host so that our tests can connect to the database.
    - _GitHub configures the container network to enable communication from the service container to the Docker host._ (https://docs.github.com/en/free-pro-team@latest/actions/guides/about-service-containers)
-  how the caching works
   - We need to cache because for jobs and workflows that are very similar (e.g. developer A and developer B trigger CI build), we don't want to install all our dependencies from a clean slate.
   - Instead, we can cache the yarn cache directory so subsequent jobs and workflows can reuse the cache. The cache is regenerated if we detect the OS has changed or the `yarn.lock` file has changed
   - https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows

One interesting thing I found out is that for each workflow, Github actions provisions a virtual machine. And inside the virtual machine, we run our jobs, and can also use containers (like Postgres) inside the VM. I'm not sure if I got the concept completely correct, pls correct me if I got something wrong.

Things we need to do:
- Separate BE and FE into their own repositories (don't want to run this workflow for PRs that are only FE)
- Have a conversation about other CI services
  - If anyone has experience with other CI services (CircleCI, Jenkins, Travis, etc), I would love to learn more about your experience with them. I've worked with Jenkins before, but have not set it up and found it really helpful. I'm open to completely scrapping this and using another CI service or just to get to know your experiences with them.

Open to any feedback and willing to answer any questions! The documentation for Github Actions is very thorough and easy to follow too.


backend/package.json
```diff
@@ -8,7 +8,8 @@
     "prestart": "yarn run build",
     "start": "nodemon src/main/App.ts",
     "test": "jest",
-    "lint": "eslint . --ext .ts,.tsx --fix",
+    "lint": "eslint . --ext .ts,.tsx",
+    "lintFix": "eslint . --ext .ts,.tsx --fix",
```
ericrkuo, line 12: @andrewzulaybar I'm not quite too familiar with this area. I've been finding the interaction between lint and prettier a bit weird at times. Sometimes lint makes us use a single quote, but then prettier changes it back to a double quote. I disabled prettier for now, not too sure what to do or what we can do.

backend/yarn.lock
```diff
@@ -2333,16 +2331,6 @@ fs.realpath@^1.0.0:
...
-fsevents@^2.1.2:
-  version "2.2.1"
-  resolved "https://registry.yarnpkg.com/fsevents/-/fsevents-2.2.1.tgz#1fb02ded2036a8ac288d507a65962bd87b97628d"
-  integrity sha512-bTLYHSeC0UH/EFXS9KqWnXuOl/wHK5Z/d+ghd5AsFMYN7wIGkUCOJyzy88+wJKkZPGON8u4Z9f6U4FdgURE9qA==
-
-fsevents@~2.1.2:
-  version "2.1.3"
-  resolved "https://registry.yarnpkg.com/fsevents/-/fsevents-2.1.3.tgz#fb738703ae8d2f9fe900c33836ddebee8b97f23e"
-  integrity sha512-Auw9a4AxqWpa9GUfj370BMPzzyncfBABW8Mab7BGWBYDj4Isgq+cDKtx0i6u9jcX9pQDnswsaaOTgTmA5pEjuQ==
-
```
ericrkuo, line 2345: I removed this because whenever I do yarn install, I always get a warning. I figured that `fsevents` is only for MacOS and I use windows. Is this important for MacOS?

andrewzulaybar (approved):
This looks great!!! Thanks for looking into this and getting this setup, this will be super helpful next semester to make sure we don't break things once we start building on top of old functionality :)

I also added an issue for separating the BE/FE into their own repos: https://github.com/ubclaunchpad/foodies/issues/91

Also, I don't know too much about the differences between the CI platforms, but I'm thinking that Github Actions may be enough for most things that we need? Have you come across any limitations of GA in the research that you did? I have some (limited) experience with CircleCI from my last co-op but haven't really worked with it to the point where I could make any comparisons... if you were looking at the other options besides GA I'd say that's a good lightweight choice :)

backend/package.json
andrewzulaybar: @ericrkuo Ah yeah I remember running into that issue before too. Prettier is just supposed to automate fixing the linting issues for you, but sometimes I think they have conflicting rules with the lint rules we set. I think a way around this is either to set the Prettier config to align with our lint rules (https://prettier.io/docs/en/configuration.html) or you could remove it like that and just fix the linting issues manually, whatever you think is easiest!

.github/workflows/foodies-backend-ci.yml
```diff
@@ -0,0 +1,156 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Foodies BE CI
+
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
```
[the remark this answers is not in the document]
ericrkuo: Yeah I was thinking that too. To me, I didn't see a point to run CI for PRs that aren't merging into master since that's the branch we care about at the end of the day.

ericrkuo:
> Also, I don't know too much about the differences between the CI platforms, but I'm thinking that Github Actions may be enough for most things that we need? Have you come across any limitations of GA in the research that you did? I have some (limited) experience with CircleCI from my last co-op but haven't really worked with it to the point where I could make any comparisons... if you were looking at the other options besides GA I'd say that's a good lightweight choice :)

Some limitations I've noticed is that there's a lot of duplicated steps in our yaml file, but Github actions doesn't have ways to remove that duplication. I haven't researched the limitations of GA yet my bad 😅 . ohhh I see, sure I'll take a look into CircleCI

backend/package.json
ericrkuo: Ohh okay let me take a look to see if I can add that in

backend/yarn.lock
andrewzulaybar: Oh interesting, what was the warning you were getting? Not sure how important for MacOS this is because I haven't come across it yet... Depending on the error it might help to clear your cache before running yarn install?

andrewzulaybar:
> Some limitations I've noticed is that there's a lot of duplicated steps in our yaml file, but Github actions doesn't have ways to remove that duplication. I haven't researched the limitations of GA yet my bad 😅 . ohhh I see, sure I'll take a look into CircleCI

Yeah makes sense, I also noticed that and I guess it would get annoying once we start adding more jobs 😅 And yeah no worries! I just remembered that we paid like $800/month for CircleCI oops so unless they have a free plan with good perks I'm not sure if it'll be worth it to set up? The only nice thing I remember is that they had a dashboard with all the currently running jobs and stuff and you can see logs to help pinpoint the cause of failure, but I think GA does a similar thing if one of your tests doesn't pass like you can see which one?

backend/yarn.lock
ericrkuo: I'd always come across this message. I tried researching this and found that its a matter of OS, but I can spend some time to research how to fix this properly
image

ericrkuo:
> Yeah makes sense, I also noticed that and I guess it would get annoying once we start adding more jobs 😅 And yeah no worries! I just remembered that we paid like $800/month for CircleCI oops so unless they have a free plan with good perks I'm not sure if it'll be worth it to set up? The only nice thing I remember is that they had a dashboard with all the currently running jobs and stuff and you can see logs to help pinpoint the cause of failure, but I think GA does a similar thing if one of your tests doesn't pass like you can see which one?

Oh yeah definitely. You can see a history of all jobs and currently running jobs here. https://github.com/ubclaunchpad/foodies/actions

And each job u can look into the console log output as well

backend/yarn.lock
andrewzulaybar: Ohhh is this the same error? https://stackoverflow.com/questions/46929196/how-to-solve-npm-install-throwing-fsevents-warning-on-non-mac-os

Is the pic above from GA? Cause I'm thinking if it is and it's running Linux, the first answer might be helpful

Either way, I don't think there should be too much harm in just removing it for now. If we figure out that there's a need for this dependency when developing locally we can deal with it then?

andrewzulaybar:
> Oh yeah definitely. You can see a history of all jobs and currently running jobs here. https://github.com/ubclaunchpad/foodies/actions
> 
> And each job u can look into the console log output as well

Ooooh honestly this looks pretty good already then! 😛

backend/yarn.lock
ericrkuo: Ohh yah same error. I get the same thing for Windows as well. Yup 100% agree, sounds like good idea

ld357 (approved):
Looks great!! Thanks for setting it up 😄  👍🏻

Issue #71, opened 2020-11-19
Backend needs unit tests for controllers


ericrkuo:
We need to mock request, response, and next function in order to test controllers and assert the behaviour is as expected. This is high priority as we continue expanding our code, we need to ensure everything works

PR #94, opened 2020-12-29
Added controller tests

ericrkuo:
https://github.com/ubclaunchpad/foodies/issues/71

Problem: We don't have integration tests for our controllers.

Approach:
- used [SuperTest](https://github.com/visionmedia/supertest) to add tests for our controllers
- refactored `App.ts` and added `Index.ts` and `BaseController.ts` to support `SuperTest`
- `Index.ts` is now responsible for running our application

Additional Stuff
- Modified `yarn run jest` command so that more is outputted in our CI logs (now we see logs for each individual test and how long it took)
https://github.com/ubclaunchpad/foodies/blob/89c049fbb32ef7899cbda6cf19ae4445eb8d8ddd/backend/package.json#L10
    - before in Github Actions CI:
image

    - after in Github Actions CI (just for one test file):
image

Now we can (hopefully) be more confident that any changes in our controllers won't break any existing or expected behaviour


backend/src/main/App.ts
```diff
@@ -28,110 +28,120 @@ import { SavedPromotion } from './entity/SavedPromotion';
...
 /* eslint-disable  @typescript-eslint/no-unused-vars */
 //todo: ormconfig.json should not have synchronize and drop schema as true for production
-createConnection()
-  .then(async (connection) => {
-    const app = express();
-    app.use(bodyParser.json());
-    const PORT = 8000;
-
-    // await loadSampleData();
+export class App {
```
ericrkuo, line 31: There are not any major changes here other than the fact that all these methods have been refactored into the class `App`. Also, `registerHandlersAndRoutes` now has the responsibility of registering handlers and routes. `init` is only responsible for creating the app and listening to the port

backend/src/main/controller/UserController.ts
```diff
@@ -89,7 +89,7 @@ export class UserController {
           UserRepository
         );
         const result = await userRepository.save(user);
-        return res.status(201).send(result);
+        return res.status(201).send({ ...result, password: undefined });
```
ericrkuo, line 92: @yunishimura0716 I made some changes to `UserController.ts` and `UserRouter.ts`. Are these changes okay?

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -0,0 +1,288 @@
...
+    request(app)
+      .get('/promotions')
+      .expect(200)
+      .end((err, res) => {
+        if (err) return done(err);
+        const promotions = res.body;
+        expect(promotions).toHaveLength(1);
+        comparePromotions(promotions[0], promotion);
+        done();
+      });
```
ericrkuo, line 56: This is `supertest` syntax, notice how we pass in `app` to `request`.

ericrkuo (commented):
Open to feedback!

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -0,0 +1,288 @@
...
+    const schedule = new ScheduleFactory().generate();
+    const expectedPromotion = new PromotionFactory().generate(user, discount, [
+      schedule,
+    ]);
+
+    await userRepository.save(user);
+    await promotionRepository.save(expectedPromotion);
+
+    request(app)
+      .get(`/promotions/${expectedPromotion.id}`)
```
ericrkuo, line 160: Example GET request with url parameters

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -0,0 +1,288 @@
...
+    const expectedPromotion = new PromotionFactory().generate(
+      user,
+      new DiscountFactory().generate(DiscountType.PERCENTAGE),
+      [new ScheduleFactory().generate()]
+    );
+
+    await userRepository.save(user);
+    request(app)
+      .post('/promotions')
+      .send({ ...expectedPromotion, user: undefined, userId: user.id })
```
ericrkuo, line 181: Example POST request sending with request body

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -0,0 +1,288 @@
...
+    await userRepository.save(user);
+    await promotionRepository.save(promotion1);
+    await promotionRepository.save(promotion2);
+    await promotionRepository.save(promotion3);
+
+    request(app)
+      .get('/promotions')
+      .query({
+        searchQuery: searchKey,
+      })
```
ericrkuo, line 132: Example GET request with query params

backend/src/main/controller/UserController.ts
yunishimura0716: What does this line `92` do ? Does this line hide the actual password when users create their account?

yunishimura0716:
Test files for controllers are so readable that makes sense to me. I'm just wondering that when we create PR, all test files we tested by jest will be running?

backend/src/main/controller/UserController.ts
ericrkuo: yeah I didn't see much of a point returning the password or even the encrypted password when we create a user. Although I think once we add our user auth login (with firebase potentially), we probably don't need this anymore

ericrkuo:
> Test files for controllers are so readable that makes sense to me. I'm just wondering that when we create PR, all test files we tested by jest will be running on CI?

Yup CI will run all the tests in our `test` directory whenever we create a PR that wants to merge into master, or we push a commit into that PR. It actually calls this command here:
https://github.com/ubclaunchpad/foodies/blob/89c049fbb32ef7899cbda6cf19ae4445eb8d8ddd/.github/workflows/foodies-backend-ci.yml#L153-L156

And `yarn run test` is equivalent to
https://github.com/ubclaunchpad/foodies/blob/89c049fbb32ef7899cbda6cf19ae4445eb8d8ddd/backend/package.json#L10

yunishimura0716 (approved):

ericrkuo:
Strange, I got this deadlock error for the previous run. I only was able to reproduce this once locally and I ran all the tests about 20+ times.

I saw a Github issue for TypeORM how it might be related to caching (two transactions trying to update the cache at the same time, but both waiting on one another).

Going to keep rerunning the jobs here and see how often it occurs.

image

backend/src/main/Index.ts
```diff
@@ -0,0 +1,4 @@
+import { App } from './App';
+
+const app: App = new App();
+app.init().then();
```
andrewzulaybar, line 4: Ahh, curious -- is there a difference without the empty `.then` block?

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -0,0 +1,339 @@
...
+    ]);
+
+    await userRepository.save(user);
+    await promotionRepository.save(promotion);
+
+    request(app)
+      .get('/promotions')
+      .expect(200)
+      .end((err, res) => {
+        if (err) return done(err);
```
andrewzulaybar, line 51: Does this statement `done(err)` fail the test if executed?

backend/src/test/controller/PromotionController.test.ts
```diff
@@ -0,0 +1,339 @@
...
+      .query({
+        discountType: DiscountType.PERCENTAGE,
+      })
+      .expect(200)
+      .end((err, res) => {
+        if (err) return done(err);
+        const promotions = res.body;
+        expect(promotions).toHaveLength(1);
+        comparePromotions(promotions[0], promotion1);
+        done();
```
andrewzulaybar, line 94: Also just wondering, is the difference in this format vs. awaiting the call to `request(app)...expect(200)` just a stylistic choice?

andrewzulaybar (approved):
These look great, nice job :) Just left some questions for my own understanding!

backend/src/main/Index.ts
ericrkuo: Oh should be no difference, I was getting like a warning that the promise returned is ignored so I just added it

backend/src/test/controller/PromotionController.test.ts
ericrkuo: Yup!

backend/src/test/controller/PromotionController.test.ts
ericrkuo: Ohh yeah, actually that's a good point I didn't realize thanks for bringing it up.
I noticed I kind of mix. 
```
await request(app).get(...)
// do something after

vs

request(app).get(...)
   .end((err, res) => {
       // do something after
   })
```

I can make this more consistent actually. I'm kind of thinking of removing all the repository calls in the `// do something after` sections because of the problem mentioned here https://github.com/ubclaunchpad/foodies/pull/94#issuecomment-752329183 (not the deadlock issue, but how the transaction hasn't committed and how our repository calls would not get up to date information).

backend/src/test/controller/UserController.test.ts
```diff
@@ -0,0 +1,398 @@
...
+            );
+            const user = await userRepository.findOneOrFail({
+              id: expectedUser.id,
+            });
+            expect(user).toMatchObject(changedProperties);
+            done();
+          }
+        );
+      });
+  });
```
ericrkuo, line 132: Ok so I think this will solve one of the issues where we were reading outdated data.
The problem was we issue PATCH request. PATCH request starts a transaction and makes the changes. We call `userRepository.find` and the data hasn't changed. Then the transaction from the PATCH request commits.  Hence we were reading not up to date data.

I think we can fix this by making sure that any repository interaction after issuing supertest request is under a transaction that has isolation level `READ UNCOMMITTED`. So we do not need to depend on the PATCH request to commit.

We still have the problem about the deadlocks, can't seem to figure that out since it occurs very rarely. But I do think we need to prepare our code for deadlocks (e.g. if a deadlock occurs, rollback the current transaction and try again after X seconds)

ericrkuo: @andrewzulaybar are you okay with me merging still even though sometimes we may face the deadlock issue here? https://github.com/ubclaunchpad/foodies/pull/94#issuecomment-752329183. I think I fixed the other ones

ericrkuo: Weird I can't reproduce the deadlock anymore....i've tried about 30 times locally LOL

ericrkuo: I'll proceed with merging. If we keep encountering this problem a lot of times we can bug it and dedicate some further time to look into it!

Issue #74, opened 2020-11-20
Backend: Filter promotions by schedule


ericrkuo:
As a user, I want to know which promotions are available on a specific day or specific time

Implementaiton details:
Will need to modify PromotionQueryValidation and PromotionQueryValidationDTO

ericrkuo:
Not doing time, not until we have a design for it

PR #103, opened 2021-01-02
74 - Filter promotions by schedule

ericrkuo:
https://github.com/ubclaunchpad/foodies/issues/74

We can now query promotions by a specific day of the week

Also exposing Day enum as endpoint for this PR


backend/src/main/repository/PromotionRepository.ts
```diff
@@ -71,6 +72,24 @@ export class PromotionRepository extends Repository<Promotion> {
...
+            .from(Schedule, 'S')
+            .where('S.dayOfWeek = :dayOfWeek')
+            .getQuery();
+          return 'promotion.id in ' + subQuery;
+        },
+        {
+          dayOfWeek: promotionQuery.dayOfWeek,
+        }
+      );
+    }
```
ericrkuo, line 91: This translates to SQL query like. The reason we want to use a subquery is that we still want to return all the schedules associated with a promotion
```
select * from promotion 
inner join discount on promotion.id = discount."promotionId"
inner join schedule on promotion.id = schedule."promotionId"
where promotion.id in (select S."promotionId" from schedule S where S."dayOfWeek" = :dayOfWeek);
```

andrewzulaybar (approved):
This looks good to me! Nice tests and thanks for getting this set up so quickly :)

On a somewhat related note, do you think the BE could also expose the `Day` enum as an endpoint? I think we could make use of it on the FE too!

ericrkuo:
> This looks good to me! Nice tests and thanks for getting this set up so quickly :)
> 
> On a somewhat related note, do you think the BE could also expose the `Day` enum as an endpoint? I think we could make use of it on the FE too!

Ah yup I can make that change in this PR!

backend/src/test/repository/EntityIntegration.test.ts
```diff
@@ -664,4 +665,63 @@ describe('Integration tests for all entities', function () {
...
+
+      expect(promotions).toBeDefined();
+      expect(promotions.length).toEqual(2);
+      const expectedPromotions = [promotion1, promotion3];
+      for (const expectedPromotion of expectedPromotions) {
+        const p = promotions.find(
+          (promotion) => promotion.id === expectedPromotion.id
+        );
+        expect(p).toBeDefined();
+        expect(p!.schedules.length).toEqual(expectedPromotion.schedules.length);
```
ld357, line 716: Just curious what the '!' notation means? :)

ld357 (approved):
Looks good 💯

backend/src/test/repository/EntityIntegration.test.ts
ericrkuo: Oh its the bang operator. Basically it tells the compiler I know it will not be undefined or null when I reach that line of code. Because on the previous line, I have an `expect` that checks that p is defined, but the compiler isn't smart enough to know that.
image

PR #106, opened 2021-01-03
Query promotions that fit into any of specified cuisines

ericrkuo:
Problem: Currently, we can only query promotions by one cuisine type. We want to be able to find all promotions that fit under any arbitrary number of cuisine types (for example I want to know all promotions that have cuisine types Chinese, Korean, and Japanese.

Approach:
When we call `GET /promotions`, the controller can now handle an array of Cuisine Types or just a string Cuisine Type

image
- By following the format in the image (from https://stackoverflow.com/a/33086861), express will automatically convert it into an array


backend/src/main/validation/PromotionQueryValidation.ts
```diff
@@ -16,7 +16,11 @@ export class PromotionQueryValidation {
      * */
     discountValue: Joi.number().positive().precision(2),
     promotionType: Joi.string().valid(...Object.values(PromotionType)),
-    cuisine: Joi.string().valid(...Object.values(CuisineType)),
+    /** request.query needs to follow format in order to be automatically converted to array - https://stackoverflow.com/a/33086861 */
+    cuisine: Joi.alternatives(
+      Joi.string().valid(...Object.values(CuisineType)),
+      Joi.array().items(Joi.string().valid(...Object.values(CuisineType)))
+    ),
```
ericrkuo, line 23: @andrewzulaybar is it difficult to format the request query in the image? I got this from here https://stackoverflow.com/a/33086861. 
image

I found this an easier solution because if I made the request query something like this I would need to use `JSON.parse` and it got complicated with Joi
image

backend/src/main/repository/PromotionRepository.ts
```diff
@@ -40,9 +40,16 @@ export class PromotionRepository extends Repository<Promotion> {
...
+      if (Array.isArray(promotionQuery.cuisine)) {
+        // https://github.com/typeorm/typeorm/issues/1239#issuecomment-366955628
+        queryBuilder.andWhere('promotion.cuisine in (:...cuisine)', {
+          cuisine: promotionQuery.cuisine,
+        });
+      } else {
+        queryBuilder.andWhere('promotion.cuisine = :cuisine', {
+          cuisine: promotionQuery.cuisine,
+        });
+      }
```
andrewzulaybar, line 52: Pretty minor optimization, I think you can assign `promotionQuery.cuisine` to a variable before this if-else block and make use of it throughout cause it looks like there are 3 usages

backend/src/main/validation/PromotionQueryValidation.ts
andrewzulaybar: I don't think that'd be too hard, so I can change it!

andrewzulaybar (approved):
Nice tests!!! I just made a note to myself to make the change you requested in the PR I'm currently working on but besides that, it looks good!

backend/src/main/repository/PromotionRepository.ts
ericrkuo: Oh, do you mind explaining how it optimizes the performance? Or do you mean like for readability

backend/src/main/validation/PromotionQueryValidation.ts
ericrkuo: Sweet thanks!

engineereng (approved):
Looks good to me!

ld357 (approved):
Looks good! Maybe an extra test case if the string is empty or the array is empty? Just wondering what would happen for those scenarios in our code!

ericrkuo:
> Looks good! Maybe an extra test case if the string is empty or the array is empty? Just wondering what would happen for those scenarios in our code!

Oh if the string is empty then we get a validation error from Joi because it's not a valid cuisine type.  This is an error reported at the controller level, but I didn't test this case inside `EntityIntegrationTest` because the `PromotionQueryDTO` will only allow `CuisineType | CuisineType[]`

In terms of the empty array that's a great question! I tested it out and that seems to throw an error, so I made the modifications to prevent against an empty array. Good eye, thanks for catching that!

PR #107, opened 2021-01-04
Fix docker compose file and delete ormconfig.json

ericrkuo:
I found it kind of troublesome that I had to keep switching back and forth depending on whether we wanted to use docker-compose or not. Because in `ormconfig.json` we need to specify `host` as `db` if we use docker-compose, and if we want to develop locally, we need to specify the host as `localhost`. To fix this, I believe we can use environment variables! This will also be better once we have a deployed version of our DB server and can then change the host.

There's not really anything mandatory you need to do, but it's recommended to do the following:
- create a duplicate of `sample.env` and rename to `.env` (shouldn't need to modify anything in `sample.env`)

Testing I did
- made sure the app still runs with or without docker-compose


yunishimura0716 (approved):

andrewzulaybar (approved):
Yeah, this seems like a good solution. Feel free to merge!

Issue #109, opened 2021-01-06
Docker-compose


ericrkuo:
Currently when we run `docker-compose up -d` it runs yarn install. However, I find that after I use docker-compose, I need to run yarn install all over again if I want to use the application without docker-compose. 

I think when we use docker-compose, it somehow changes all the dependencies in the node modules of our repo. One theory we can test is to see if `yarn install --frozen-lockfile --prefer-offline` helps. Another theory is that we might need to make changes in the Dockerfile for the backend repo. But regardless, the first step is to research to see if others have the same issue.

PR #110, opened 2021-01-08
Fix docker compose file

ericrkuo:
https://github.com/ubclaunchpad/foodies/issues/109

I had to run `yarn install` all over again when switching from `docker-compose` to developing without `docker-compose`. I think this is because of what's mentioned below...I'm not too sure but all I know is the problem is not there anymore.

I also noticed that when I used `docker-compose` it would take a long time to start, only for the `foodies_web` container. I realized this is because we specified this command to run:
image
But actually, we don't need to install dependencies again. The docker image will be responsible for installing dependencies, not when we run the containers. Hence, I believe we can just do `yarn run start (or yarn start)`. 

I moved the command into the `Dockerfile` instead because compose will run the command inside the `Dockerfile` when we execute the container
image

Now using docker-compose is much faster


backend/Dockerfile
```diff
@@ -1,8 +1,9 @@
 FROM node:15.0.1
 WORKDIR /foodies/backend
-COPY *.json /foodies/backend/
+COPY package.json ./
+COPY yarn.lock ./
```
ericrkuo, line 4: Added to copy `yarn.lock`, no need for wildcard because we may end up copying `eslintrc.json` which we do not need for `RUN yarn install`.

I also checked that the dest `./` works because it's relative to `WORKDIR`. I verified by viewing the file system in the docker container and saw that it was copied correctly

docker-compose.yml
```diff
@@ -25,7 +25,6 @@ services:
           - ./backend:/foodies/backend
         ports:
           - "8000:8000"
-        command: bash -c "yarn install && yarn start"
         depends_on: 
           - db
         links:
```
ericrkuo, line 30: ALso in case you're wondering, in our docker-compose file we specify the environment vars here 
image
When we build the docker image, one of the steps in our Dockerfile says `COPY . .` so we still end up copying the dotenv file. And the dotenv file also has the environment var `DB_HOST`. However, the environment var we specified in the compose file take precedence over the ones specified in the `dotenv` file.

Source: https://vsupalov.com/override-docker-compose-dot-env/

yunishimura0716 (approved):
Thank you for improving the docker-compose file. Just add one thing. You can run `yarn run test` after you get in the docker container by `docker-compose exec web bash`.

ericrkuo:
> Thank you for improving the docker-compose file. Just add one thing. You can run `yarn run test` after you get in the docker container by `docker-compose exec web bash`.

Ah I see, but we would need to modify the Postgres container to have the database `foodies_test` rather than `foodies` right?

README.md
```diff
@@ -17,4 +17,12 @@ Load the back-end in a container:
 ```
 docker-compose exec web bash
 ```
+Stop docker compose without removing containers
```
ld357, line 20: Nice job updating the docs again!

ld357 (approved):
Thanks for fixing this!! 🎉

PR #116, opened 2021-02-02
Updated Multiple Library Packages

ld357:
Resolves this issue: https://github.com/ubclaunchpad/PromoPal-backend/issues/102

I upgraded a couple packages including ESLint, Prettier, Husky, and Lint-Staged, etc to the suggested versions in the warning column provided by running `yarn outdated`.  Now, when you run `yarn run lintFix` there should be no warnings.


ericrkuo (approved):
Looks good! Do all the commands in package.json work still and not output anything funky?

ld357:
> Looks good! Do all the other commands in package.json work still and not output anything funky?

Everything's working except for the `typeorm` command- it doesn't seem to be a valid command because I'm getting suggested commands?

yunishimura0716:
Thank you for your work @ld357 ! So, before we push the branch, we should do `yarn run lintFix`, right?

ericrkuo:
> > Looks good! Do all the other commands in package.json work still and not output anything funky?
> 
> Everything's working except for the `typeorm` command- it doesn't seem to be a valid command because I'm getting suggested commands?

I looked into this cause I forgot what the `typeorm` command is for 😅 . it's used for sync schema and drop schema commands, so by itself it doesn't do anything.
image

ericrkuo:
> Thank you for your work @ld357 ! So, before we push the branch, we should do `yarn run lintFix`, right?

@yunishimura0716 Oh `yarn run lintFix` was complaining before because we upgraded TypeScript dependency and the eslint plugin didn't support the current version of TypeScript. So we needed to upgrade eslint and now it doesn't give us a warning message. There's nothing we need to do before we push the branch I think, because we have pre-commit hooks to run lint and prettier, and CI also checks linting as well.

andrewzulaybar (approved):
Thanks for looking into this @ld357 and @ericrkuo for catching it! Feel free to merge when you're ready

PR #117, opened 2021-02-07
Update README.md

richardyjtian:


README.md
```diff
@@ -2,22 +2,25 @@
...
 
 ## Configure environment variables
 
+Make a copy of `sample.env`, rename it to `.env`, and fill out the environment variables respectively.
+They're likely to be:
 ```
-Make a copy of `sample.env` and rename it to `.env`.
-Fill out the environment variables respectively
+DB_HOST=localhost:5432
+REDIS_HOST=localhost:6379
```
ericrkuo, line 23: Oh strange, you needed to specify the ports for the environment variables?
Mine just looks like this atm
image

README.md
```diff
@@ -42,50 +45,53 @@ GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO postgres;
...
 ALTER USER postgres CREATEDB;
 
-// repeat these two lines for both databases
 CREATE DATABASE foodies;
 
 GRANT ALL PRIVILEGES ON DATABASE foodies TO postgres;
+
+CREATE DATABASE foodies_test;
+
+GRANT ALL PRIVILEGES ON DATABASE foodies_test TO postgres;
```
ericrkuo, line 54: Thanks for this, I was lazy :P

ericrkuo (approved):
Thanks for updating read me, looks much cleaner and easier to understand now

README.md
```diff
@@ -2,22 +2,25 @@
...
 
-1. Postgres
+1. Yarn (https://classic.yarnpkg.com/en/docs/install/)
+1. Postgres (https://www.postgresql.org/download/)
    - when installing, use the default password "postgres"
    - for macOS users, suggested installation is through homebrew
 2. TypeORM Global Installation (`yarn add typeorm -g`)
    - this will install the TypeORM CLI as well
-3. Redis
+3. Redis (https://redis.io/download)
```
ld357, line 11: Thanks for adding this 😅  Do you mind changing to "Local Redis Instance"?

README.md
```diff
@@ -2,22 +2,25 @@
 
 ## Requirements
 
-1. Postgres
+1. Yarn (https://classic.yarnpkg.com/en/docs/install/)
+1. Postgres (https://www.postgresql.org/download/)
```
ld357, line 6: Minor thing but numbering is off!

README.md
ld357: Mine works like that too ^!

ld357 (changes requested):
Thanks for adding the fixes!! 🥳

ld357:
Could you also add this command `docker system prune -a --volumes` ? It removes all containers, volumes, networks and images if you want to restart everything before running `docker-compose up -d --build` I usually get random bugs/errors with docker-compose but this always saves the day!

README.md
ericrkuo: How come "Local Redis Instance"?

README.md
richardyjtian: hmm maybe I didn't need to specify the port then haha

ericrkuo (approved):

andrewzulaybar (approved):
Looks good to me too! Thanks for adding these details in :)

README.md
ld357: It's just to be clear that it's being used locally- since the deployment one will be the Docker version

ld357 (approved):
Thanks for addressing the changes, looks good to me! :)

README.md
ericrkuo: Hmm I'm not sure I understand, usually for local development it's suggested that the services you use are local instances. Deployment is a whole other story right

ld357: Yeah! Like the setup you follow here is so that you can get a local Redis instance running instead of using the Redis with Docker for development

ericrkuo: Oh, wait rather than `deployment`, you meant development with Docker? Yah docker development is covered in other section of the ReadMe

https://github.com/ubclaunchpad/PromoPal-backend/blob/7e3a4b421efc01e8161f37df7e5d3558cf9db9a4/README.md#local-development-with-docker

Otherwise, I think we would need to say "Local Postgres Instance" then

ld357: Ahh when I said deployment, I meant that the Redis we'll use when our app deployed, is the Redis associated with Docker! And in that case, I think we should follow "Local Postgres Instance" as well- so future devs don't get mixed up

ericrkuo:
Hey @richardyjtian, feel free to merge this PR in. Seems like there are some conflicting files. I was thinking in the future I might just get rid of `init_data.sql` since I'm assuming not many people use that to load the sample data into the DB.

PR #118, opened 2021-02-11
PP-10 Addressed NOT_FOUND Error in GooglePlacesService & Missing Place ID in Caching Service

ld357:
**PART ONE:**
We may encounter the NOT_FOUND error when we issue a Places Detail request for a specific restaurant. This will require us to make a Place Details - ID Refresh SKU request with only the ID field in the parameter.

We may encounter NOT_FOUND from the refresh request as well- which means we will need to get the restaurant name and location from the BE and make a Places Search again through getRestaurantInfo. This part is not addressed in this PR- it will be wherever we're making the call to fetch info for a particular restaurant.

**The changes for this PR is just adding a column for `restaurantLocation`.**

[LINK TO FRONT-END PR](https://github.com/ubclaunchpad/PromoPal-frontend/pull/16) 

**PART TWO:**
When a `placeID` is missing from the `CachingService`, an empty string will be returned instead. The particular promotion(s) associated with this `placeID` that are returned either through `GET /promotion` or `GET /promotions` will not have any lat/lon values. This will be handled in the FE as such: https://promopal.atlassian.net/browse/PP-19?focusedCommentId=10000


src/main/entity/Promotion.ts
```diff
@@ -168,6 +170,13 @@ export class Promotion {
   @Column()
   restaurantName: string;
 
+  /*
+   * Represents restaurant location from Google Places API
+   * Many promotions can come from the same restaurant and thus have the same restaurantLocation
+   * */
+  @Column()
+  restaurantLocation: string;
+
```
ericrkuo, line 179: How about `restaurantAddress`?

Also now that I think about it, notice how we have a lot of `restaurant___` properties now inside the Promotion entity. Maybe we should make this a one to one association and make a `Restaurant` entity.

src/main/entity/Promotion.ts
```diff
@@ -168,6 +170,13 @@ export class Promotion {
   @Column()
   restaurantName: string;
 
+  /*
+   * Represents restaurant location from Google Places API
+   * Many promotions can come from the same restaurant and thus have the same restaurantLocation
```
ericrkuo, line 175: We're not allowed to store this though right?

src/main/entity/Promotion.ts
richardyjtian: Yea, I think we can consider creating a new Restaurant entity in the future, but two properties doesn't seem so bad right now.

richardyjtian (approved):

andrewzulaybar (approved):

src/main/entity/Promotion.ts
ld357: We'll be storing the user inputted version! We need this in order to refresh the invalid place ID on the FE

ericrkuo: Ohya but the comment says its from Google places api, could we update that?

ericrkuo (commented):
Also could we update the promoting controller integration tests to check for this new field?

src/main/entity/Promotion.ts
ericrkuo: Alrighty sounds good, is the name restaurantAddress okay though rather than restaurantLocation?

src/main/entity/Promotion.ts
ld357: Ooh I kept it as "Google Places API" cause last time I changed `restaurantName`'s description to something along "inputted by user"- you mentioned that it was giving away too much FE implementation details?

src/main/entity/Promotion.ts
ld357: Yup I can change it!

src/main/entity/Promotion.ts
ericrkuo: Hmm not sure where I implied that, this point in ur previous PR was for discussion https://github.com/ubclaunchpad/PromoPal-backend/pull/81#discussion_r551717463. If I did, sorry that was my mistake.

I realized we can't store a restaurants name or address from google places API. So then `restaurantName` and `restaurantAddress` don't represent anything from google places API

ericrkuo: How about 
`The restaurant address of the promotion`

ld357: Hmmm, that wasn't the comment I was referring to-I believe you made the point in chat instead-but sure, I can change it to that

ericrkuo: My bad, but does it make sense how saying google places API doesn't really represent what this property is?

ld357: Yup! I thought it was weird when we kept it as that last time so sounds good to me!

ericrkuo (approved):
Thanks for the fixes!

src/main/service/CachingService.ts
```diff
@@ -51,13 +51,12 @@ export class CachingService {
...
    */
   async getLatLonValue(placeId: string): Promise<CachingObject> {
-    return new Promise((resolve, reject) => {
+    return new Promise((resolve) => {
       this.client.get(placeId, (err: Error | null, data: string | null) => {
-        if (err) {
-          // todo: https://github.com/ubclaunchpad/foodies/issues/89
-          reject(err);
+        if (err || !data) {
+          return resolve(JSON.parse('{}'));
```
ericrkuo, line 57: This is equivalent right?

```suggestion
        if (err || !data) {
          return resolve({});
```

ericrkuo, line 57: Wait how come this is resolved? 

If there is no more lat lon in the cache for a specific placeId, we would never get the lat lon again unless we call `POST /promotions`

src/test/controller/PromotionController.test.ts
```diff
@@ -404,6 +404,7 @@ describe('Unit tests for PromotionController', function () {
       startDate: expectedPromotion.startDate.toISOString(),
       lat: expectedPromotion.lat,
       lon: expectedPromotion.lon,
+      restaurantAddress: expectedPromotion.restaurantAddress,
```
andrewzulaybar, line 407: Should `restaurantName` be here as well?

src/main/service/CachingService.ts
ld357: I'm gonna keep the old version cause Typescript isn't accepting it!

`Argument of type '{}' is not assignable to parameter of type 'CachingObject | PromiseLike<CachingObject>'.   Property 'then' is missing in type '{}' but required in type 'PromiseLike<CachingObject>'`

ericrkuo: Hmm ok generally if any language is complaining about types, then we should try to properly fix it. I think the JSON.parse is a "hack" that should be avoided

PR #120, opened 2021-02-13
Abstract conversion from DTO to entity

ericrkuo:
The logic of converting DTO's to entity's kind of clouded our logic inside the controller. I abstracted this behaviour out and created a class `DTOConverter.ts` to convert DTO's to entity's. This will help make our application code cleaner as we expand on the number of entity's and DTO's we have


src/main/controller/PromotionController.ts
```diff
@@ -153,8 +126,6 @@ export class PromotionController {
           promotionDTO.lat,
           promotionDTO.lon
         );
-        result.lat = promotionDTO.lat;
-        result.lon = promotionDTO.lon;
```
ericrkuo, line 157: Don't need this because `DTOConverter.promotionDTOToPromotion` will set the lat lon for the `promotion`. And when we save this using the repository, even though lat lon aren't columns, `result` will still contain lat lon

src/main/validation/PromotionValidation.ts
```diff
@@ -53,7 +52,7 @@ export class PromotionValidation {
 export interface PromotionDTO {
   userId: string;
   discount: DiscountDTO;
-  schedules: Schedule[];
+  schedules: ScheduleDTO[];
```
ericrkuo, line 55: Made a sad mistake here ):

ld357: Good catch though! 😄

src/main/validation/DTOConverter.ts
```diff
@@ -0,0 +1,77 @@
...
+import { UserDTO } from './UserValidation';
+
+/**
+ * Converts DTOs to entity's
+ * */
+export class DTOConverter {
+  /**
+   * Convert PromotionDTO to a Promotion
+   */
+  static promotionDTOtoPromotion(
```
ld357, line 17: Just curious but is there a particular reason why you chose static fns?

ld357 (commented):
Nice abstraction! Just a few questions :)

src/main/controller/PromotionController.ts
ericrkuo: Oh it doesn't get saved because lat/lon aren't columns

src/main/validation/DTOConverter.ts
ericrkuo: Good question! This class doesn't really have any "state" to maintain so I didn't really see a point in instantiating it and injecting it into `PromotionController` and `UserController`. We could've just removed the whole class, and exported each as a function, but yah I didn't give much thought to it.

ld357: Oooh cool I see!! That makes sense

ld357 (approved):
Thanks for the update!! :))

richardyjtian (approved):

PR #121, opened 2021-02-19
Use pessmistic writes to avoid deadlocking for controller integration tests

ericrkuo:
Our controller integration tests for UserController would deadlock for the `savedPromotions` tests.
 
It would usually deadlock when we do the following
1. Save User
2. Save promotion
3. Save saved promotion <-- could deadlock at this step
4. GET or DELETE saved promotion using supertest <-- could deadlock at this step

Approach
- to solve deadlocking at step 3, I made 1,2,3 execute in separate transactions that would have lock level `PESSMISTIC_WRITE`.
- to solve 4...unfortunately I could not wrap my head around this, so I opted for a sad timeout of 0.5 seconds 😞. 

My knowledge of deadlocks and in general locking is not great. Feel free to suggest any changes and if you've had experience with locking/deadlocks would appreciate any help.


ericrkuo:
@yunishimura0716 try pasting this code into your current PR, this is a rough attempt at trying to solve the deadlock errors. If it works, I'll explain in more detail in this PR what the changes are.

ericrkuo:
Nevermind, seems like there's a deadlock ):

yunishimura0716:
> Nevermind, seems like there's a deadlock ):

So, you mean that you're still working to fix the deadlock?

ericrkuo:
> > Nevermind, seems like there's a deadlock ):
> 
> So, you mean that you're still working to fix the deadlock?

Yeah, sorry for the false alarm I thought it was working. Will resume looking into it today or tmr

yunishimura0716:
> > > Nevermind, seems like there's a deadlock ):
> > 
> > 
> > So, you mean that you're still working to fix the deadlock?
> 
> Yeah, sorry for the false alarm I thought it was working. Will resume looking into it today or tmr

I think my implementation of protected endpoint for user_profile might be helpful to see the deadlock, so please refer it. (`user-endopoints` branch)

src/main/controller/UserController.ts
```diff
@@ -156,15 +158,31 @@ export class UserController {
...
-        res.status(200).send(savedPromotions);
+        let promotions: Promotion[] = [];
+
+        if (promotionIds.length !== 0) {
+          // get all promotions using the promotion id's, DO NOT join to discount/schedules, we don't need all the information
+          promotions = await transactionalEntityManager
+            .getCustomRepository(PromotionRepository)
+            .find({
+              id: In(promotionIds),
+            });
```
ericrkuo, line 182: Changed the contract for getting a user's saved promotions. I didn't see the point of returning information about the user, so I just returned an array of promotions. I also don't think we need to join to discount or schedules, don't think that information is necessary for viewing a user's saved promotions.

yunishimura0716:
I pull this branch and run the tests with firebase auth. However, I still got the deadlock.

yunishimura0716 (approved):

src/test/controller/UserController.test.ts
```diff
@@ -399,4 +375,59 @@ describe('Unit tests for UserController', function () {
 
     expect(actualUser).toMatchObject(expectedObject);
   }
+
+  /**
+   * Save the promotion and user to the database. Also add the saved promotion into the database
+   * * Set transaction level to serializable and set lock level to pessimistic write to avoid deadlocks
+   * */
+  async function addSavedPromotion(user: User, promotion?: Promotion) {
```
richardyjtian, line 383: I think the deadlock may be occurring because we're trying to do too much in one function. In this case, we should have a function specifically for creating a user, and another function for creating a promotion and a saved promotion. In the second function, reading the user would not require a lock, and we can simply abort the transaction if the user gets deleted during this time.

src/main/controller/UserController.ts
ericrkuo: Just kidding, I think we do need to return discount or schedules as well...will fix this

src/test/controller/UserController.test.ts
ericrkuo: Hmm I'm not too sure I understand how moving it to separate functions will solve the deadlock. Unless you mean one transaction for saving the user and another transaction for creating a promotion and a saved promotion.

Right now there doesn't seem to be any deadlocks for saving a user, promotion, and saved promotion. The deadlock seems to happen when we send the request to the controller after saving these entities.

src/test/controller/UserController.test.ts
```diff
@@ -399,4 +375,59 @@ describe('Unit tests for UserController', function () {
...
+                userId: user.id,
+              },
+            ])
+            .execute();
+        }
+      );
+    }
+
+    // This is an unfortunate hack...need to invest more time into this
+    await new Promise((resolve) => setTimeout(resolve, 500));
```
andrewzulaybar, line 431: Sorry, maybe I'm missing this but where do you do step (4) here? i.e. GET or DELETE saved promotion

Is it the `connection.clear()`?

andrewzulaybar (approved):

src/test/controller/UserController.test.ts
ericrkuo: Oh step (4) is in the actual test case itself when we use `supertest` to test our endpoints

src/main/controller/UserController.ts
ericrkuo: Okok I think to be safe I'll just join everything again

PR #122, opened 2021-02-21
Implementation and refactoring of protected user endpoints by firebase auth

yunishimura0716:
I set up firebase admin sdk on backend and implemented the authorization middleware to protect the endpoint to access user profile. Also, refactored not to use password on our Backend.

The task of middleware is simple. Take the `IdToken` on the request header, then decode it to verify. If the `IdToken` is what appropriately created in FE side, the middleware allow the request to go to next processing. 

To use firebase auth on production, we need `FIREBASE_SERVICE_ACCOUNT_ID`, `FIREBASE_PROJECT_ID`, `FIREBASE_PRIVATE_KEY`, `FIREBASE_CLIENT_EMAIL` for environment variable.

The config file of firebase is at under `service` directory.

For tests, I was using `firebase-mock` library which make mock firebase admin to create user and authentication.

The issues I did not figure out is to fix the deadlock problem happened in user endpoint tests.
Also, we need to ensure the user on firebase and our DB are synchronous (if user exist in firebase server, the user mush have own info in our DB). I did not implement this functionality yet. This would need to consider the design of FrontEnd because FE should progress `SignUp` and `Delete` with our DB.


package.json
```diff
@@ -30,13 +30,17 @@
...
     "bcryptjs": "^2.4.3",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
+    "firebase-admin": "^9.4.2",
+    "firebase-mock": "^2.3.2",
     "joi": "^17.3.0",
     "redis": "^3.0.2",
-    "redis-mock": "^0.56.3"
+    "redis-mock": "^0.56.3",
+    "request-promise": "^4.2.6"
```
ericrkuo, line 43: I'm seeing some unmet peer dependency messages, I think all you need to do is install these dependencies and the messages will go away

image

src/main/service/FirebaseConfig.ts
```diff
@@ -0,0 +1,15 @@
...
+  admin.initializeApp({
+    serviceAccountId: process.env.FIREBASE_SERVICE_ACCOUNT_ID,
+    credential: admin.credential.cert({
+      projectId: process.env.FIREBASE_PROJECT_ID,
+      privateKey: process.env.FIREBASE_PRIVATE_KEY,
+      clientEmail: process.env.FIREBASE_CLIENT_EMAIL,
+    }),
+  });
+  return admin;
+}
```
ericrkuo, line 15:
```suggestion
export function initFirebaseAdmin(): Auth {
  admin.initializeApp({
    serviceAccountId: process.env.FIREBASE_SERVICE_ACCOUNT_ID,
    credential: admin.credential.cert({
      projectId: process.env.FIREBASE_PROJECT_ID,
      privateKey: process.env.FIREBASE_PRIVATE_KEY,
      clientEmail: process.env.FIREBASE_CLIENT_EMAIL,
    }),
  });
  return admin.auth();
}
```

How about this? That way everything has types now and we don't need `any`

src/main/service/FirebaseConfig.ts
```diff
@@ -0,0 +1,15 @@
...
+import * as dotenv from 'dotenv';
+dotenv.config();
+
+export function initFirebaseAdmin() {
+  admin.initializeApp({
+    serviceAccountId: process.env.FIREBASE_SERVICE_ACCOUNT_ID,
+    credential: admin.credential.cert({
+      projectId: process.env.FIREBASE_PROJECT_ID,
+      privateKey: process.env.FIREBASE_PRIVATE_KEY,
+      clientEmail: process.env.FIREBASE_CLIENT_EMAIL,
```
ericrkuo, line 11: Also, could we add all these new environment variables to `sample.env` and update the read me too?

src/test/controller/BaseController.ts
```diff
@@ -1,21 +1,55 @@
...
   return redisMock.createClient();
 };
+
+export const createFirebaseMock = () => {
+  const mockauth = new firebasemock.MockAuthentication();
+  const mockdatabase = new firebasemock.MockFirebase();
+  const mockfirestore = new firebasemock.MockFirestore();
+  const mockstorage = new firebasemock.MockStorage();
+  const mockmessaging = new firebasemock.MockMessaging();
+  const mocksdk = new firebasemock.MockFirebaseSdk(
```
ericrkuo, line 32: Could we change all the variable names to lower camel case?

src/main/App.ts
```diff
@@ -60,7 +67,8 @@ export class App {
    * */
   async registerHandlersAndRoutes(
     app: Express,
-    redisClient: RedisClient
+    redisClient: RedisClient,
+    firebaseadmin: any
```
ericrkuo, line 71:
```suggestion
    redisClient: RedisClient,
    firebaseAdmin: Auth
```

I think we can make this lower camel case and also make it typed

package.json
```diff
@@ -30,13 +30,17 @@
     "@types/bcryptjs": "^2.4.2",
     "@types/redis": "^2.8.28",
     "@types/redis-mock": "^0.17.0",
-    "axios": "^0.21.0",
+    "@types/request-promise": "^4.1.47",
+    "axios": "^0.21.1",
     "bcryptjs": "^2.4.3",
```
ericrkuo, line 35: I think we can also get rid of `bcrypt.js` dependency now

src/test/controller/BaseController.ts
```diff
@@ -1,21 +1,55 @@
 import express, { Express } from 'express';
 import { App } from '../../main/App';
 import redisMock, { RedisClient } from 'redis-mock';
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const firebasemock = require('firebase-mock');
```
ericrkuo, line 5: Unfortunately seems like `firebase-mock` doesn't have types for typescript ): hmm let's see if we can find a mock that comes with type definitions

src/test/controller/UserController.test.ts
```diff
@@ -363,12 +422,13 @@ describe('Unit tests for UserController', function () {
       });
   });
 
-  test('GET /users/:id/uploadedPromotions - should return empty list if user has no uploaded promotions', async (done) => {
+  xtest('GET /users/:id/uploadedPromotions - should return empty list if user has no uploaded promotions', async (done) => {
```
ericrkuo, line 425: You're getting deadlocks for uploadedPromotions endpoints? I was only seeing deadlocks for savedPromotions

ericrkuo (commented):
Good job so far, I've only reviewed half so far, my comments are mainly about type safety. Usually, we want to refrain from using `any` because we want to detect type errors at compilation time and to leverage type safety

src/main/service/FirebaseConfig.ts
ericrkuo: Then also for `FirebaseAuth` it would look something like this
image

yunishimura0716: Good refactoring! thanks!

src/test/controller/BaseController.ts
yunishimura0716: I'll refactor these!

src/main/service/FirebaseConfig.ts
yunishimura0716: Hmm, I think we could not since the web api key and private key of firebase must be secret.

src/test/controller/UserController.test.ts
yunishimura0716: This is because the deadlock happened once on this function, but it would be fixed so changed to test

src/main/App.ts
yunishimura0716: I refactored that, thanks

src/main/middleware/Authentication.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Middleware function to check if the request has correct client header using fireabse-admin auth
+ */
+export class FirebaseAuth {
+  private admin: Auth;
+
+  constructor(firebaseAdmin: Auth) {
+    this.admin = firebaseAdmin;
+  }
+
+  isAuthorizedForNewUser = async (
```
ld357, line 16: This function seems pretty similiar to `isAuthorizedForProtection` - is there a reason you separated the two? Is it possible to capture the duplication into one function?

src/main/controller/UserController.ts
```diff
@@ -83,7 +84,7 @@ export class UserController {
           UserRepository
         );
         const result = await userRepository.save(user);
-        return res.status(201).send({ ...result, password: undefined });
+        return res.status(201).send({ ...result, idFirebase: undefined });
```
ld357, line 87: Just curious but is there a reason why we're sending back `idFirebase` if we'll be setting it to undefined?

src/main/controller/UserController.ts
```diff
@@ -156,15 +154,31 @@ export class UserController {
...
         );
-        const savedPromotions = await userRepository.findOneOrFail(id, {
-          relations: ['savedPromotions', 'savedPromotions.promotion'],
-          cache: true,
-        });
 
-        res.status(200).send(savedPromotions);
+        let promotions: Promotion[] = [];
+
+        if (promotionIds.length !== 0) {
```
ld357, line 172: I believe you can simplify check to `if (promotionIds.length)`

src/main/middleware/Authentication.ts
```diff
@@ -0,0 +1,56 @@
...
+    res: Response,
+    next: NextFunction
+  ): Promise<any> => {
+    const idToken: string = req.headers.authorization!;
+
+    try {
+      const decodedToken = await this.admin.verifyIdToken(idToken);
+
+      if (decodedToken) {
+        req.body.idFirebase = decodedToken.uid;
```
ld357, line 27: Minor suggestion- but maybe use `firebaseId` instead of `idFirebase` since you followed `firebaseAdmin` and `firebaseAuth`?

src/main/validation/UserValidation.ts
```diff
@@ -9,7 +9,7 @@ export class UserValidation {
     firstName: Joi.string().required(),
     lastName: Joi.string().required(),
     email: Joi.string().email().required(),
-    password: Joi.string().min(8).required(),
+    idFirebase: Joi.string().required(),
```
ld357, line 12: Not sure if this would be the place to do it, but when the user is setting the password, we should probably restrict certain values (ie. 12345678) and set a character minimum

src/test/controller/BaseController.ts
```diff
@@ -1,21 +1,57 @@
...
 };
+
+export const createFirebaseMock = (): Auth => {
+  const mockAuth = new firebaseMock.MockAuthentication();
+  const mockDatabase = new firebaseMock.MockFirebase();
+  const mockFirestore = new firebaseMock.MockFirestore();
+  const mockStorage = new firebaseMock.MockStorage();
+  const mockMessaging = new firebaseMock.MockMessaging();
+  const mockSdk = new firebaseMock.MockFirebaseSdk(
+    // use null if your code does not use RTDB
```
ld357, line 35: Not sure if we aren't using any of these, but if you wanted to keep it for developers, maybe summarize all these "use null..." comments into one?

src/test/controller/EnumController.test.ts
```diff
@@ -11,11 +15,14 @@ import { RedisClient } from 'redis-mock';
 describe('Unit tests for PromotionController', function () {
   let app: Express;
   let redisClient: RedisClient;
+  let admin: any;
```
ld357, line 18: Is there a more specific type here?

src/test/controller/PromotionController.test.ts
```diff
@@ -20,13 +24,16 @@ describe('Unit tests for PromotionController', function () {
...
+  let admin: any;
   let cachingService: CachingService;
 
   beforeAll(async () => {
     await connection.create();
     mockRedisClient = await connectRedisClient();
     cachingService = new CachingService(mockRedisClient);
-    app = await registerTestApplication(mockRedisClient);
+    // init mock firebase
+    admin = createFirebaseMock();
```
ld357, line 35: Can we call this something more specific, maybe `mockFirebaseAdmin` ?

src/test/controller/UserController.test.ts
```diff
@@ -36,29 +58,46 @@ describe('Unit tests for UserController', function () {
...
+
     request(app)
       .get('/users')
+      .set('Authorization', idToken)
       .expect(200)
       .end((err, res) => {
         if (err) return done(err);
         const users = res.body;
+        // eslint-disable-next-line no-console
+        // console.log(users);
```
ld357, line 75: Remove console log!

src/test/controller/UserController.test.ts
```diff
@@ -195,52 +245,43 @@ describe('Unit tests for UserController', function () {
...
-        if (err) return done(err);
-        const user = res.body;
-        expect(user).toHaveProperty('savedPromotions');
-        compareUsers(user, expectedUser);
-        expect(user.savedPromotions).toHaveLength(1);
-        expect(user.savedPromotions[0]).toMatchObject({
-          userId: expectedUser.id,
-          promotionId: promotion.id,
-          promotion: { id: promotion.id, name: promotion.name },
+  xtest('GET /users/:id/savedPromotions', async () => {
```
ld357, line 248: should this be `test` instead of `xtest`?

src/test/controller/UserController.test.ts
```diff
@@ -195,52 +245,43 @@ describe('Unit tests for UserController', function () {
...
+    for (let i = 0; i < 100; i++) {
```
ld357, line 249: Is there a reason we're running this 100 times? 😅

src/test/validation/DTOConverter.test.ts
```diff
@@ -75,7 +75,7 @@ describe('Unit tests for DTOConverter', function () {
     expect(user).toBeInstanceOf(User);
 
     // PP-29 workaround sync we use bycrypt to hash password
-    userDTO.password = user.password;
+    // userDTO.password = user.password;
```
ld357, line 78: Can we delete this or are we keeping it for a particular reason?

ld357 (commented):
Really great job, Yu!! I didn't finish a full review yet either but mainly minor comments about spelling and clarifications! I accidentally might've added some comments that are related to the deadlock issue instead so you can ignore those since I'm assuming those are just used to solve the issue atm! 😅

src/test/controller/UserController.test.ts
ericrkuo: Ah sorry this was me loool this test is gonna take a really long time. @yunishimura0716 sorry I didn't know you were going to merge in my branch `user-endpoints-deadlock`, that branch was just me playing with things to test the deadlock. Once I merge in https://github.com/ubclaunchpad/PromoPal-backend/pull/121, then your branch should have the proper code

package.json
```diff
@@ -26,17 +26,23 @@
...
   "dependencies": {
+    "@firebase/app-types": "0.x",
     "@types/bcrypt": "^3.0.0",
     "@types/bcryptjs": "^2.4.2",
     "@types/redis": "^2.8.28",
     "@types/redis-mock": "^0.17.0",
-    "axios": "^0.21.0",
-    "bcryptjs": "^2.4.3",
+    "@types/request-promise": "^4.1.47",
+    "@firebase/app-types": "0.6.1",
```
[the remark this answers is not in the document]
yunishimura0716: hmm, I added "@firebase/app-types": "0.x" cuz I found package dependency issue. Maybe I incorrectly add the two, so it could be removed.

src/main/middleware/Authentication.ts
yunishimura0716: This is because I want to avoid setting `uid` to request body except to when making new user. Other endpoints do not have to have or should not contain `uid` in request body, so I made two different authorization functions in order to keep the same functionalities as much as possible.

src/main/controller/UserController.ts
yunishimura0716: just like password, BE do not have to send `uid` to FE since FE should get `uid` from current user of firebase.

src/main/middleware/Authentication.ts
yunishimura0716: Yeah it's better, thanks. I refactored.

src/main/validation/UserValidation.ts
yunishimura0716: I removed the validation for password since we use firebase server to store password, BE are supposed to not receive password anymore.

src/test/controller/BaseController.ts
yunishimura0716: This is the part I'm not really understanding, I just followed the way of setting up of firebase mock based on the document.

src/test/controller/EnumController.test.ts
yunishimura0716: I haven't found the good type expression for firebase mock auth. When I set `Auth` defined in `firebase admin sdk`, it does not fit.

package.json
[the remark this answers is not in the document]
ld357: Oooh, if it complained about package dependency, then it should be fine! Interesting though- never heard about needing duplicates 🤔

src/main/middleware/Authentication.ts
ld357: Hmm, would it be possible to set a flag somehow as a parameter to indicate that this is a call with making a new user?

src/main/controller/UserController.ts
ld357: Oooh could we just remove idFirebase completely?

src/main/validation/UserValidation.ts
ld357: Ahh gotcha! For the validation we have on the firebase server, what validation do we have?

src/test/controller/BaseController.ts
ld357: Do you mind pasting the link? I could take a look! :)

src/main/middleware/Authentication.ts
ericrkuo: I don't think we can do a flag for a parameter because these are middle ware functions, I think they need to have those specific parameters. 

@yunishimura0716 For the `isAuthorizedForNewUser`, won't the request body already contain `idFirebase` when we call `POST /promotions`? So then we don't need to set the `idFirebase` in the request body? https://github.com/ubclaunchpad/PromoPal-backend/blob/973b08e20c38c72acdd771da4f30c73ef8ef15ff/src/main/middleware/Authentication.ts#L27

image

src/main/service/FirebaseConfig.ts
ericrkuo: Oh I meant this file here https://github.com/ubclaunchpad/PromoPal-backend/blob/user-endpoints/sample.env. That way future devs know which api keys they need to copy. You don't need to put the actual value of the key, u can just set the value as `INSERT_HERE`.

src/main/middleware/Authentication.ts
yunishimura0716: @ericrkuo I did not implement for protected endpoint for `promotion`, but yeah, that's right. Because we can get `uid` (idFirebase here) once the `idToken` verified correctly, I want to set the `idFirebase` after success of authorization. However, if you think there's nothing problem to set `uid` at FE to request body, we can simplify the middleware function.

ericrkuo: Oops sorry, I meant `POST /users`, fixed the comment above. Is it safe to send the firebase id in the request body?

src/main/controller/UserController.ts
yunishimura0716: it just removes idFirebase from response data. I think there's no huge effect on our app even if the response data contains `idFirebase`. However, I want to just make sure that we can get `uid` when we signed up/in correctly. Every time FE requests some data to BE, the current user would be supposed to be authorized. So, when it creates new user on our DB, the user in FE can get `uid` by itself that means BE does not have to contain `idFirebase` in the response data.

src/main/middleware/Authentication.ts
yunishimura0716: I guess yes. The reason why the request body doesn't contain `uid` is because just want to make sure they can get `uid` after the `idToken` is verified. However, the `idToken` is provided from FE whose current user is signed up/in. That means, `uid` in request body should be correct and we do not have to get `uid` in BE.

src/main/validation/UserValidation.ts
yunishimura0716: sorry what do you mean the validation on the firebase server? If this means the validation for `password` and `email` when we sign up/in, I guess the validation will be done in FE since we would sign up/in between FE and firebase.

src/test/controller/BaseController.ts
yunishimura0716: This is the link for firebase mock library.
https://github.com/dmurvihill/firebase-mock

src/main/middleware/Authentication.ts
ericrkuo: Ahh okay, yes then in that case I think we should be good. Because if we were able to successfully verify the `idToken`, then we can assume the `firebaseId` in the request body is valid. I think we can simplify the middleware to be all the same then, would that be okay?

yunishimura0716: sure! I'll refactor it!

ericrkuo: Awesome thanks Yu!

src/main/service/FirebaseConfig.ts
yunishimura0716: ahh, I see lol. I'll put some needed variables.

src/main/App.ts
```diff
@@ -26,20 +26,29 @@ import { Schedule } from './entity/Schedule';
 import { SavedPromotion } from './entity/SavedPromotion';
 import redis, { RedisClient } from 'redis';
 import { CachingService } from './service/CachingService';
+import { initFirebaseAdmin } from './service/FirebaseConfig';
```
ericrkuo, line 29: I don't think `FirebaseConfig` should be in the service directory. I think usually any classes inside the service directories are classes we want to inject through constructor and perform some logic like `CachingService`. Could we move `FirebaseConfig` one level up? It will be in same directory as `App.ts` and `Index.ts`

src/main/entity/User.ts
```diff
@@ -14,20 +13,25 @@ export class User {
     lastName: string,
     email: string,
     username: string,
-    password: string
+    idFirebase: string
```
ericrkuo, line 16: What do you think about:
```suggestion
    firebaseId: string
```

src/main/middleware/Authentication.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Middleware function to check if the request has correct client header using firebase-admin auth
+ */
+export class FirebaseAuth {
+  private admin: Auth;
+
+  constructor(firebaseAdmin: Auth) {
+    this.admin = firebaseAdmin;
+  }
+
+  isAuthorizedForProtection = async (
```
ericrkuo, line 16: Nice! Really like how simple it looks now. Could we add some documentation for this function?

src/main/middleware/Authentication.ts
```diff
@@ -0,0 +1,35 @@
+import { NextFunction } from 'express';
+import { Request, Response } from 'express';
+import { auth } from 'firebase-admin/lib/auth';
+import Auth = auth.Auth;
+
+/**
+ * Middleware function to check if the request has correct client header using firebase-admin auth
+ */
+export class FirebaseAuth {
```
ericrkuo, line 9: I noticed the filename is `Authentication` whereas the class is `FirebaseAuth`. Is there a reason for making the two different?

src/test/repository/SavedPromotion.test.ts
```diff
@@ -0,0 +1,125 @@
...
+import { User } from '../../main/entity/User';
+import { Promotion } from '../../main/entity/Promotion';
+import { UserFactory } from '../factory/UserFactory';
+import { PromotionFactory } from '../factory/PromotionFactory';
+import { Discount } from '../../main/entity/Discount';
+import { DiscountFactory } from '../factory/DiscountFactory';
+import { Schedule } from '../../main/entity/Schedule';
+import { ScheduleFactory } from '../factory/ScheduleFactory';
+
+describe('Unit tests for SavedPromotionRepository', function () {
```
ericrkuo, line 15: Thanks for making these tests

src/test/controller/UserController.test.ts
```diff
@@ -1,30 +1,52 @@
...
+    app = await registerTestApplication(redisClient, mockFirebaseAdmin);
+
+    mockFirebaseAdmin.autoFlush();
+
+    // create user
+    const user = await mockFirebaseAdmin.createUser({
+      uid: uid,
+      email: 'test@gmail.com',
+      password: 'testpassword',
+    });
```
ericrkuo, line 48: Ah so what's happening here is essentially we are doing the `User sign up` logic in the FE, similar to your doc?

ericrkuo (commented):
Great job! I'm going to look into alternatives for the mock firebase, if we can't find a better mocking library let's stick with this one and I'll approve the PR. 

The reason why the current mocking library isn't good is that it seems like it's only supported for javascript, not typescript.

yunishimura0716:
> Great job! I'm going to look into alternatives for the mock firebase, if we can't find a better mocking library let's stick with this one and I'll approve the PR.
> 
> The reason why the current mocking library isn't good is that it seems like it's only supported for javascript, not typescript.

Yeah, that's true. I also will see another library but I could not find yet.

src/main/App.ts
yunishimura0716: Sure!

src/main/entity/User.ts
yunishimura0716: I refactored!

src/main/middleware/Authentication.ts
yunishimura0716: No, I refactored!

src/test/controller/UserController.test.ts
yunishimura0716: yes, actually, FE sign up/in using email and password.

src/main/route/UserRouter.ts
```diff
@@ -1,34 +1,66 @@
...
-    this.userRouter.get('/', this.userController.listAll);
-    this.userRouter.get('/:id', this.userController.getOneById);
-    this.userRouter.post('/', this.userController.newUser);
-    this.userRouter.patch('/:id', this.userController.editUser);
-    this.userRouter.delete('/:id', this.userController.deleteUser);
+    this.userRouter.get(
+      '/',
+      this.firebaseAuth.isAuthorizedForProtection,
+      this.userController.listAll
+    );
```
ericrkuo, line 23: Will we add middleware to `PromotionRouter`? We may want to authenticate the user for endpoints like `POST promotions` for example

sample.env
```diff
@@ -1,2 +1,7 @@
 DB_HOST=localhost
-REDIS_HOST=localhost
\ No newline at end of file
+REDIS_HOST=localhost
+FIREBASE_PROJECT_ID=INSERT_HERE
+FIREBASE_PRIVATE_KEY=INSERT_HERE
+FIREBASE_CLIENT_EMAIL=INSERT_HERE
+FIREBASE_WEB_API_KEY=INSERT_HERE
+FIREBASE_SERVICE_ACCOUNT_ID=INSERT_HERE
```
andrewzulaybar, line 7: Do you think you can create a file in this Google Drive folder with these keys so that we can share it with the team?
https://drive.google.com/drive/u/0/folders/1YhX35Vk8I0_PECrk6I2btsFPdd8BC-ow

src/main/entity/User.ts
andrewzulaybar: I think you should change it to `firebaseId` in this function too!

src/main/repository/UserRepository.ts
```diff
@@ -9,6 +9,10 @@ export class UserRepository extends Repository<User> {
     return this.findOne({ firstName, lastName });
   }
 
+  findByFirebaseId(idFirebase: string): Promise<User | undefined> {
+    return this.findOne({ firebaseId: idFirebase });
```
andrewzulaybar, line 13:
```suggestion
  findByFirebaseId(firebaseId: string): Promise<User | undefined> {
    return this.findOne({ firebaseId });
```

`{ firebaseId }` is shorthand for `{ firebaseId: firebaseId }`

src/main/route/UserRouter.ts
andrewzulaybar: Yeah I think this is a good idea too, probably something you can do in another PR!

src/test/controller/BaseController.ts
```diff
@@ -1,21 +1,57 @@
...
+      return mockFirestore;
+    },
+    // use null if your code does not use STORAGE
+    () => {
+      return mockStorage;
+    },
+    // use null if your code does not use MESSAGING
+    () => {
+      return mockMessaging;
+    }
```
andrewzulaybar, line 54: Oh I think what they mean here is that if we don't use firebase messaging (which I think we don't use), then we can replace this with `null`. And then we'd do the same thing for the ones above

src/test/controller/UserController.test.ts
```diff
@@ -384,19 +400,75 @@ describe('Unit tests for UserController', function () {
    * Compare actual user against expected user
    * */
   function compareUsers(actualUser: User, expectedUser: User) {
-    const expectedObject: any = { ...expectedUser };
+    const expectedObject: any = {
```
andrewzulaybar, line 403: Do you get errors when you type `expectedObject` as a User?

```suggestion
    const expectedObject: User = {
```

src/test/repository/SavedPromotion.test.ts
andrewzulaybar: I agree, this is great!

src/test/repository/SavedPromotion.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    const promotion: Promotion = new PromotionFactory().generate(
+      user,
+      discount,
+      [schedule]
+    );
+    await userRepository.save(user);
+    await promotionRepository.save(promotion);
+
+    // save the promotion
+    const saved_user: User = new UserFactory().generate();
```
andrewzulaybar, line 48: Very minor, but for code style I think you can change this variable to `savedUser`

src/test/repository/SavedPromotion.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    const schedule: Schedule = new ScheduleFactory().generate();
+    const promotion: Promotion = new PromotionFactory().generate(
+      user,
+      discount,
+      [schedule]
+    );
+    await userRepository.save(user);
+    await promotionRepository.save(promotion);
+    // save the promotion
+    const saved_user: User = new UserFactory().generate();
```
andrewzulaybar, line 77: Same as above

src/test/repository/SavedPromotion.test.ts
```diff
@@ -0,0 +1,125 @@
...
```
andrewzulaybar, line 103: Same as above

andrewzulaybar (approved):
This is really great work!!! I'm excited to get this in and to start working with it. I just made a couple minor comments if you want to take a look :)

sample.env
yunishimura0716: sure! I'll do it!

src/main/middleware/Authentication.ts
yunishimura0716: I added!

src/main/route/UserRouter.ts
yunishimura0716: We will need to discuss which endpoint we have to set the middleware.

src/test/controller/UserController.test.ts
yunishimura0716: Yes, it throws error when deleting operation.

src/test/controller/BaseController.ts
yunishimura0716: yeah, you are right. I fixed the parts.

src/test/controller/BaseController.ts
ericrkuo: @yunishimura0716 do you think this one would work https://github.com/mindhivefi/ts-mock-firebase

yunishimura0716: @ericrkuo The error occurred when I add `ts-mock-fireabse` because of build error for `gyp` and `node-pre-gyp` so that I could not install the package properly.

src/test/controller/UserController.test.ts
andrewzulaybar: Oh I see, yeah it's fine to keep it as `any` then!

src/test/controller/BaseController.ts
ericrkuo: Do you have the build output?

yunishimura0716: This is the error thrown when I tried to install `ts-mock-firebase`
Screen Shot 2021-03-02 at 12 01 58 AM

ericrkuo: Ah I see, okay let's forget about that mock library. I'm okay w using this current mocking library

ericrkuo (approved):

PR #123, opened 2021-02-27
PP-77 refactor google place behaviour

ericrkuo:
https://promopal.atlassian.net/browse/PP-77

Lily, Andrew and I discussed the interaction with Google Places and our FE and BE. We saw that there were a couple of issues with coupling and some complex logic that could be simplified.

I would highly encourage to take a look at the first two pages of https://docs.google.com/document/d/1mx9piwHnu8RCeKOoTKDRPSX61MGrNK6vzAfI0qHLCBc/edit. The interaction with Google Places is one of the core features of our app, so would be good to familiarize ourselves with how we are using google places.

What this PR covers
- Created a new entity `Restaurant` which contains the restaurant's `lat, lon`.
- We've figured out a way to store the lat/lon in our DB. Therefore, don't need caching service anymore.
- The `PlaceDetails` request now happens in the BE
- Handle the `NOT_FOUND` case of the `PlaceDetails` in the BE as well. Please let me know if you don't know what the `NOT_FOUND` case means
- We are no longer storing restaurant `name` or `address` in the DB, this is because we will be using the autocomplete feature for google maps api.

This PR is quite huge, but let me know if there's any way I can help to make it easier to understand


package.json
```diff
@@ -26,6 +26,7 @@
   },
   "homepage": "https://github.com/ubclaunchpad/foodies#readme",
   "dependencies": {
+    "@googlemaps/google-maps-services-js": "3.1.16",
```
ericrkuo, line 29: This is a great library https://googlemaps.github.io/google-maps-services-js/index.html. It's basically a wrapper around axios, but the good thing about it is **types**. We don't need to create our own interfaces.

src/main/resources/init_data.sql
```diff
@@ -1,127 +0,0 @@
-INSERT INTO public.user_profile (id, first_name, last_name, email, username, password) VALUES ('8f8fc016-5bb2-4906-ad88-68932c438665', 'John', 'Smith', 'smith.j@sample.com', 'user1', '$2a$08$6PEeGWWVF1M8hU8Y8XCNfe/FAs4WGHXX/wB0znhC3O3TxYtWY/xxe');
-INSERT INTO public.user_profile (id, first_name, last_name, email, username, password) VALUES ('66c34e55-9dec-483f-8458-1fc50e802415', 'Asa', 'Edward', 'edward.a@sample.com', 'user2', '$2a$08$gAG7ItEHUDaigRlfi9xreeK3zdbYvOsqF6x0Ic8gZpyVgK6d3j9Ba');
```
ericrkuo, line 2: Didn't see a point in keeping `init_data.sql` anymore. I don't think people use it that much, but correct me of I'm wrong. I just think loading data through `Data.ts` is much easier to maintain

src/test/mock/CustomAxiosMockAdapter.ts
```diff
@@ -0,0 +1,429 @@
...
+      params: {
+        place_id,
+        key: undefined,
+        fields: restaurantDetailsFields,
+      },
+    }).reply(
+      200,
+      //region Mocked PlaceDetailsResponseData
+      {
+        html_attributions: [],
```
ericrkuo, line 27: Ignore this gigantic sample data when reviewing

src/test/mock/CustomAxiosMockAdapter.ts
```diff
@@ -0,0 +1,429 @@
...
+import { Status } from '@googlemaps/google-maps-services-js';
+import { restaurantDetailsFields } from '../../main/service/GooglePlaceService';
+import { PlaceField } from '../../main/data/PlaceField';
+
+/**
+ * Wrapper class around MockAdapter to set custom mocking behaviour
+ * * To mock axios requests, add a new function and implement the expected response. Then in your test, call the function
+ * that was implemented to configure the mock for the respective axios instance.
+ * */
+export class CustomAxiosMockAdapter extends MockAdapter {
```
ericrkuo, line 12: We are using [axios-mock-adapter](https://www.npmjs.com/package/axios-mock-adapter) to mock [google-maps-service-js](https://googlemaps.github.io/google-maps-services-js/index.html). 

There's is no mocking library for google-maps-service. But the good thing about google-maps-service is that its constructor takes in an `axios` instance. So we can mock the `axios` instance which is realy neat!

src/test/service/GooglePlaceService.test.ts
```diff
@@ -0,0 +1,109 @@
...
+      const placeDetailsResponseData: PlaceDetailsResponseData = await googlePlaceService.getRestaurantDetails(
+        SAMPLE_PLACE_ID
+      );
+      expect(placeDetailsResponseData.status).toEqual('OK');
+      expect(placeDetailsResponseData.result.place_id).toEqual(SAMPLE_PLACE_ID);
+      expect(placeDetailsResponseData.result.name).toEqual('MOCK NAME');
+    } catch (e) {
+      fail('Should not have failed: ' + e);
+    }
+  });
```
ericrkuo, line 41: Example of using mocks in our test

src/test/service/GooglePlaceService.test.ts
```diff
@@ -0,0 +1,109 @@
...
+ * */
+describe('Unit tests for GooglePlaceService', function () {
+  const SAMPLE_PLACE_ID = 'ChIJb0n5cWl3hlQRIbVGYLiTEgE';
+  let googlePlaceService: GooglePlaceService;
+  let customAxiosMockAdapter: CustomAxiosMockAdapter;
+
+  beforeEach(() => {
+    const axiosInstance = axios.create();
+    googlePlaceService = new GooglePlaceService(new Client({ axiosInstance }));
+    customAxiosMockAdapter = new CustomAxiosMockAdapter(axiosInstance);
```
ericrkuo, line 26: Example of using the Axios instance in `Client` and `CustomAxiosMockAdapter`

src/test/mock/CustomAxiosMockAdapter.ts
```diff
@@ -0,0 +1,429 @@
...
+      //endregion
+    );
+  }
+
+  /**
+   * Mock invalid request response for place Details
+   * * Note: Invalid Request responses are still status code 200
+   * */
+  mockInvalidRequestPlaceDetails(): void {
+    this.onGet(defaultUrl).reply(200, {
```
andrewzulaybar, line 365: I'm wondering for this response and the not found response, do they also get returned with status 200?

src/main/route/PromotionRouter.ts
```diff
@@ -24,6 +24,10 @@ export class PromotionRouter {
       '/:id/downVote',
       this.promotionController.downVotePromotion
     );
+    this.promotionRouter.get(
+      '/:id/restaurantDetails/',
+      this.promotionController.getRestaurantDetails
+    );
```
andrewzulaybar, line 30: Just confirming, this is the endpoint we'll call to get the restaurant details of a certain promotion?

In src/main/repository/PromotionRepository.ts, it looks like we get all the restaurant details along with the promotions so I was wondering if this was needed by some other part of the app? If not, I think it's fine to keep supporting this but I'm just wondering if we have a current use case on the FE

src/main/service/GooglePlaceService.ts
```diff
@@ -0,0 +1,101 @@
...
+    return refreshResult;
+  }
+}
+
+/**
+ * Refresh result when refreshing placeId
+ * @property placeId the new placeId
+ * @property restaurantDetails the result of place details request with the new placeId, or undefined if placeId refresh unsuccessful
+ * */
+interface RefreshResult {
```
andrewzulaybar, line 98: I think it would be nice to have the type declarations at the top of the file!

src/test/controller/PromotionController.test.ts
```diff
@@ -377,17 +313,109 @@ describe('Unit tests for PromotionController', function () {
...
   });
 
-  async function cacheLatLonForPromotions(promotions: Promotion[]) {
-    for (const promotion of promotions) {
-      if (promotion.lat && promotion.lon) {
-        await cachingService.cacheLatLonValues(
-          promotion.placeId,
-          promotion.lat,
-          promotion.lon
+  test('GET /promotions/:id/restaurantDetails/:placeId', async (done) => {
```
andrewzulaybar, line 316: Should this be
```suggestion
  test('GET /promotions/:id/restaurantDetails', async (done) => {
```

and same for the other test titles?

andrewzulaybar (commented):
This is really great work @ericrkuo! Thanks for working on this and also for working with @ld357 on coming up with the high-level design. Just had a couple questions here!

src/main/controller/UserController.ts
```diff
@@ -181,7 +181,7 @@ export class UserController {
               where: {
                 id: In(promotionIds),
               },
-              relations: ['discount', 'schedules'],
+              relations: ['discount', 'schedules', 'restaurant'],
```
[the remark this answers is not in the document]
ericrkuo: Hmm, I was thinking I might as well return everything, including the restaurant (which now only entails lat/lon). No specific reason but I was thinking of keeping it consistent with GET/promotions and GET/promotions/:id

src/main/entity/Promotion.ts
```diff
@@ -82,6 +81,16 @@ export class Promotion {
...
 
+  /*
+   * OneToOne bidirectional relationship between Promotion and Restaurant
+   * Each promotion can have a single restaurant
+   * */
+  @OneToOne(() => Restaurant, (restaurant) => restaurant.promotion, {
+    cascade: true,
+    nullable: false,
+  })
+  restaurant: Restaurant;
```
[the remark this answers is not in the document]
ericrkuo: Omg okay that gave me a new idea,
I can do a `ManyToOne` relationship with
```
Restaurant
   id:
   placeId:
   lat:
   lon:
```

When creating a new promotion, we try to find a restaurant with the same placeId first, otherwise we create a new restaurant (thus new placeId). 

I think I'll need to test the cascade deletes such that:
If many promotions reference the same promotion, if you delete a promotion it won't delete the restaurant. But if you delete all the promotions, it would delete the restaurant...

src/test/mock/CustomAxiosMockAdapter.ts
ericrkuo: Correct!

src/main/route/PromotionRouter.ts
ericrkuo: Oh the one in promotionrepository does not contain the data back from google places api, that one contains just the Restaurant entity which is lat lon. To get the restaurant details itself, we need to make this separate request.

I'm not sure if we'll run into rate limiting issues if we get all promotions along with their restaurant details from places API. By having this separate, we only make this request when a user clicks on restaurant card

src/main/service/GooglePlaceService.ts
ericrkuo: Got it, i'll keep this in mind for the future thanks

src/test/controller/PromotionController.test.ts
ericrkuo: Yes good catch, I must have missed this change when I changed the endpoint. Thanks!

src/main/controller/PromotionController.ts
```diff
@@ -204,4 +196,70 @@ export class PromotionController {
...
+          if (placeDetailsResponseData.status === Status.NOT_FOUND) {
+            result = await this.handlePlaceIdNotFound(
+              promotion.placeId,
+              id,
+              transactionalEntityManager
+            );
+          }
+        }
+
+        return response.status(200).send(result);
```
ld357, line 235: If the `promotion.placeId` is an empty string, we return an empty result?

src/main/controller/UserController.ts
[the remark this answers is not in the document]
ld357: I think it might be nice if we end up adding a section on the homepage where you can see your saved promotions with the map? Feature doesn't exist yet but could work (?)

src/main/data/PlaceField.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * The Place Data Fields from https://developers.google.com/places/web-service/place-data-fields.
+ * Specifies the types of Place data to return when requesting Place Details
+ * */
+export enum PlaceField {
```
ld357, line 5: Do we need all these fields?

src/main/controller/PromotionController.ts
```diff
@@ -204,4 +196,70 @@ export class PromotionController {
...
+    transactionalEntityManager: EntityManager
+  ): Promise<Place> {
+    const refreshResult = await this.googlePlaceService.refreshPlaceId(placeId);
+
+    // update DB with new placeId, even if placeId is empty string
+    await transactionalEntityManager
+      .getCustomRepository(PromotionRepository)
+      .update({ id }, { placeId: refreshResult.placeId });
+
+    return refreshResult.restaurantDetails;
```
ld357, line 263: Should we also set the placeIDs of all the other promotions associated with the same restaurant to empty as well? And also each of their associated restaurant entity?

src/main/service/GooglePlaceService.ts
```diff
@@ -0,0 +1,101 @@
+import {
+  Client,
+  Place,
+  PlaceDetailsResponse,
+} from '@googlemaps/google-maps-services-js';
+import { PlaceField } from '../data/PlaceField';
+import { PlaceDetailsResponseData } from '@googlemaps/google-maps-services-js/dist/places/details';
```
ld357, line 7: This is such a useful library!! 💯

src/main/service/GooglePlaceService.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Refresh the placeId by making a refresh request.
+   * * If there is a new placeId available, get the restaurant details
+   * and return them along with the placeId.
+   * * If refresh request does not succeed, then RefreshResult will return an empty string for placeId and no restaurant details.
+   * @param nonExistentPlaceId the placeId that resulted in NOT_FOUND which triggered this function call
+   * @return RefreshResult
+   * */
+  async refreshPlaceId(nonExistentPlaceId: string): Promise<RefreshResult> {
```
ld357, line 73: can we rename nonExistentPlaceId to invalidPlaceId?

src/test/service/GooglePlaceService.test.ts
```diff
@@ -0,0 +1,109 @@
...
+      expect(placeDetailsResponseData.status).toEqual('OK');
+      expect(placeDetailsResponseData.result.place_id).toEqual(SAMPLE_PLACE_ID);
+      expect(placeDetailsResponseData.result.name).toEqual('MOCK NAME');
+    } catch (e) {
+      fail('Should not have failed: ' + e);
+    }
+  });
+
+  test('If response results in INVALID_REQUEST, no data should be present', async () => {
+    customAxiosMockAdapter.mockInvalidRequestPlaceDetails();
```
ld357, line 44: Just to help me better understand the mock, how do these calls ensure it's using the mocked version?
ie `customAxiosMockAdapter.mockInvalidRequestPlaceDetails()`?

ld357 (commented):
Thanks for making these changes!! Looks great 🥳

src/main/controller/PromotionController.ts
ericrkuo: Yup correct. We can spend a larger effort in future to account for placeId being empty string, which only happens if refresh request fails

src/main/data/PlaceField.ts
ericrkuo: Right now, no. But in the future maybe, I think there's no harm having the complete set of fields here. Is that okay?

src/main/controller/PromotionController.ts
ericrkuo: That's a very good point, yes. I think once I have refactored it into a oneToMany relationship this issue won't exist, but that's a very good point

src/main/service/GooglePlaceService.ts
ericrkuo: 100% agree, having types is very useful and helps to detect errors at compilation. I would recommend we use the same library in the FE, except we only use it for its interfaces. This library isnt recommended for client side.

src/test/service/GooglePlaceService.test.ts
ericrkuo: good question. 

First, we need to configure the mocks. We're making sure that `Client` and `CustomAxiosMockAdapter` share the same `axiosInstance`, otherwise this wouldn't work.
https://github.com/ubclaunchpad/PromoPal-backend/blob/6886d32e9dc59757253105162479a46a6d756cfe/src/test/service/GooglePlaceService.test.ts#L23-L27

By saying `customAxiosMockAdapter.mockInvalidRequestPlaceDetails()`, we're telling the mock adapter that on any axios request that hits the specified endpoint, we will return an invalid request response. Since `Client` uses `axios` to make the request, we've mocked axios so instead of making an http request, we'll return what the mock has configured.

src/main/controller/UserController.ts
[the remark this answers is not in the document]
andrewzulaybar: Ahhh okay makes sense! But yeah @ld357 that's a cool idea, I remember that idea was brought up so maybe we will end up needing this in the future 👀

src/main/entity/Promotion.ts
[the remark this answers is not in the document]
andrewzulaybar: Ohhh yeah that sounds like a good idea, I like it! And then in the case that the placeId is NOT_FOUND, we leave it as empty string again? So then `id` would be the primary key and `placeId` would be non-unique?

And ya that's a good point, I can't think of what it would be off the top of my head... maybe to not delete on cascade? Yeah maybe double-check that 😅

src/main/route/PromotionRouter.ts
andrewzulaybar: Oh I see yeah that makes sense, thanks for explaining! Yeah this is a good solution cause we probably don't want to make those extra API calls if we don't need to

src/main/service/GooglePlaceService.ts
andrewzulaybar: Oh yeah we use `@types/googlemaps` on the FE which I think this library uses for types? I've used it a bit for the maps but I was thinking most of the Google Places types now would be abstracted by the types returned from the BE (which might be a good thing in terms of loosening the dependency between FE and Google Places)

src/main/service/GooglePlaceService.ts
ericrkuo: Actually I think I'll keep it nonExistent. Invalid can suggest other things like malformed string. Non existent to me is clear that the placeId results in a not found error

ericrkuo:
@andrewzulaybar @ld357 
I made the changes, was more of a headache than I imagined.

Now we have the following:
- a promotion has a many to one relationship with a restaurant
- a restaurant has a one to many relationship with a promotion
- **important**: Because of the new relationship b/t promotion and restaurant, `Restaurant` is the parent or reference table. `Promotion` is the referencing table or the child with the FK constraint. Because of the nature of FK constraints, if I set cascade delete on the table `Promotion`, anytime I delete a parent, it will delete the child. In other words, every time we delete a restaurant, it will delete the promotions. This is **not** what we want
    - Instead, we want the other way around. When I delete a promotion, and no other promotion is referencing the restaurant, I want to delete the restaurant. FK constraints can't work the other way around like this.
    - What I ended up doing was creating a trigger. Please take a look at this file as it is very important.
    - https://github.com/ubclaunchpad/PromoPal-backend/blob/5007454df01944cd87f239170c796264adeb964e/src/main/migration/1614748794259-RestaurantDeletion.ts#L11
- Also now when we add a promotion in the `PromotionController`, we look for existing restaurants with the same placeId.
- `Restaurant` now has the `placeId` property rather than `Promotion`. As of consequence:
   - we have `RestaurantRouter` and `RestaurantController`
   - We have a new endpoint `restaurants/:id/restaurantDetails`. This was moved out of `Promotioncontroller` because this endpoint no longer concerns or relates to a promotion.

src/main/controller/PromotionController.ts
```diff
@@ -109,24 +103,29 @@ export class PromotionController {
...
           promotionDTO,
-          user
+          user,
+          // todo PP-82 if restaurant does not exist create a new one with correct lat and lon
+          restaurant ??
+            new Restaurant(
+              promotionDTO.placeId,
+              randomLatitude(),
+              randomLongitude()
+            )
```
ericrkuo, line 122: @ld357 important part when u rebase ur branch, this is the only place I think that should conflict with ur work

src/main/migration/1614748794259-RestaurantDeletion.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * This migration is to support restaurant cleanup when a promotion is deleted.
+ * * We only want to delete a promotions restaurant if the restaurant is referencing only one more promotion
+ * * Thus create a trigger to execute on for each promotion (row) that is deleted in the promotion table. We can reference
+ * the deleted promotion through `old`. Then create a function to delete the restaurant based on the deleted promotion `old` only if
+ * the restaurant is referencing zero promotions.
+ * * Intro to triggers: https://www.postgresqltutorial.com/creating-first-trigger-postgresql/
+ * */
+export class RestaurantDeletion1614748794259 implements MigrationInterface {
```
andrewzulaybar, line 11: I looked into this briefly and came across these and was wondering if you think this would accomplish the same thing:

https://github.com/typeorm/typeorm/pull/7105
https://medium.com/platri-it/how-to-delete-orphaned-rows-full-example-with-typeorm-nestjs-and-postgresql-a8baa29c01f8

Seems like a pretty new feature they added, but it looks like it requires a higher version of typeorm

andrewzulaybar (approved):
This is amazing! Can't imagine how much effort you put into this. Thanks for working on this @ericrkuo and hopefully we don't come across any other issues with Google Places 😛

src/test/service/GooglePlaceService.test.ts
ld357: Oooh, gotcha! Thanks for explaining 😄

src/main/entity/Promotion.ts
```diff
@@ -82,13 +73,19 @@ export class Promotion {
...
+   * See RestaurantDeletion migration for how we handle this
+   *
+   * * We do NOT want cascade delete on foreign key restaurantId because we do not want to delete the promotions when a restaurant is deleted
+   * * Promotion is owning side of this association, contains column restaurantId
    * */
-  @OneToOne(() => Restaurant, (restaurant) => restaurant.promotion, {
-    cascade: true,
+  @ManyToOne(() => Restaurant, (restaurant) => restaurant.promotion, {
     nullable: false,
+    cascade: true,
```
ld357, line 86: Just wanna double check- I thought we didn't want cascade delete?

ld357 (approved):
Just had one clarifying question but really awesome work, Eric!! This looked really complicated so kudos 🥳

src/main/entity/Promotion.ts
ericrkuo: Ah yup this is tricky if you think of `cascade` in terms of cascade deletes. Actually, the cascade keyword for typeorm means if I have a promotion with a restaurant, and I save the promotion, the restaurant will be saved automatically. If I didn't have this keyword, I would need to save promotion first, then save restaurant separately

src/main/migration/1614748794259-RestaurantDeletion.ts
andrewzulaybar: Ahhh okay makes sense I got it confused 😅 Seems weird that there isn't a simpler solution cause this seems like it would be a common issue

PR #124, opened 2021-03-09
BE - Refactor Upload Promotion Google Place Behaviour - Geocoder Implementation

ld357:
Instead of using the Redis cache we had initially, Andrew, Eric and I decided to use `node-geocoder` instead for a restaurant's lat/lon values. You can read more about the reasoning in the [doc here](https://docs.google.com/document/u/1/d/1mx9piwHnu8RCeKOoTKDRPSX61MGrNK6vzAfI0qHLCBc/edit?usp=drive_web&ouid=102604531027328302119)  

Basically, when we receive the address from AutoComplete on the FE, we include the address in the PromotionDTO that gets sent during `POST /promotions`. If the restaurant does not exist in our DB, we use `node-geocoder` to get the lat/lon coordinates based on this address and then we instantiate/save this restaurant.

I put the base branch as PP-77 (feature branch) so it's easier to notice the changes! Once that's merged, I'll update to master.

I also chose to use locationIQ for this because I tested it out with other supported libraries like MapQuest and OpenStreetMap and it was the most accurate (in comparison to result provided by Google Places API)

[View Associated JIRA Ticket](https://promopal.atlassian.net/browse/PP-82?atlOrigin=eyJpIjoiNjAyYTE1NWIwODkzNGZmZTg4OTdmMGQ5OGU0Y2EyMDQiLCJwIjoiaiJ9)


src/main/controller/PromotionController.ts
```diff
@@ -103,23 +111,37 @@ export class PromotionController {
...
+            promotionDTO.googlePlacesAddress
+          );
+          if (coordinates.lat && coordinates.lon) {
+            restaurant = new Restaurant(
+              promotionDTO.placeId,
+              coordinates.lat,
+              coordinates.lon
+            );
+          } else {
+            restaurant = new Restaurant(promotionDTO.placeId);
```
ld357, line 134: Here, we'll save the restaurant regardless if we're unable to fetch lat/lon values! FE will deal with whether or not lat/lon are null

ericrkuo, line 134: maybe geocodingservice.createrestaurantfromcoordinates?

src/main/entity/Restaurant.ts
```diff
@@ -33,12 +33,18 @@ export class Restaurant {
   /**
    * Latitude coordinates of restaurant
    * */
-  @Column('real')
-  lat: number;
+  @Column({
+    type: 'real',
+    nullable: true,
+  })
+  lat: number | null;
```
ericrkuo, line 40: I think this can just be
```suggestion
  lat?: number;
```

Then the constructor doesn't need the `| null` part

src/main/service/GeocodingService.ts
```diff
@@ -0,0 +1,40 @@
+import nodeGeocoder, { Entry, Geocoder, Options } from 'node-geocoder';
+import { GeocodingObject } from '../data/GeocodingObject';
+import * as dotenv from 'dotenv';
+
+/**
+ * A Node library for geocoding our restaurant addresses
```
ericrkuo, line 6: How come we need to mention Node?

src/main/service/GeocodingService.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export class GeocodingService {
+  private geocoder: Geocoder;
+
+  constructor() {
+    dotenv.config();
```
ericrkuo, line 12: Oh we don't need to do dotenv config here, index.ts calls it once when we start our application. 

But if its needed for tests, I made comment below about tests

src/main/service/GeocodingService.ts
```diff
@@ -0,0 +1,40 @@
...
+    const options: Options = {
+      provider: 'locationiq',
+      apiKey: process.env.GEOCODING_KEY,
+    };
+    this.geocoder = nodeGeocoder(options);
+  }
+
+  public getGeoCoordinatesFromAddress(
```
ericrkuo, line 20: Some documentation for this function if you don't mind :)

Also could we use await and async, no need for catch since errors will be caught at controller level

src/main/service/GeocodingService.ts
```diff
@@ -0,0 +1,40 @@
...
+    address: string
+  ): Promise<GeocodingObject> {
+    return this.geocoder
+      .geocode(address)
+      .then((entries: Entry[]) => {
+        const result: GeocodingObject = {};
+
+        if (entries?.length) {
+          const geocodeResult = entries[0];
```
ericrkuo, line 29: So first one is always the most accurate?

src/main/service/GeocodingService.ts
```diff
@@ -0,0 +1,40 @@
...
+export class GeocodingService {
+  private geocoder: Geocoder;
+
+  constructor() {
+    dotenv.config();
+    const options: Options = {
+      provider: 'locationiq',
+      apiKey: process.env.GEOCODING_KEY,
+    };
+    this.geocoder = nodeGeocoder(options);
```
ericrkuo, line 17: Just curious, wanted to know why we went with this provider and how we decided this one was better

src/main/service/GeocodingService.ts
```diff
@@ -0,0 +1,40 @@
...
+ * A Node library for geocoding our restaurant addresses
+ */
+export class GeocodingService {
+  private geocoder: Geocoder;
+
+  constructor() {
+    dotenv.config();
+    const options: Options = {
+      provider: 'locationiq',
+      apiKey: process.env.GEOCODING_KEY,
```
ericrkuo, line 15: Do you have a link for where u signed up for api key? I'm just curious cause gonna look into the usage limit and other stuff

src/test/controller/PromotionController.test.ts
```diff
@@ -153,12 +154,13 @@ describe('Unit tests for PromotionController', function () {
     const expectedPromotion = new PromotionFactory().generateWithRelatedEntities(
       user
     );
+    const inputtedPromotion = setGooglePlacesAddress(expectedPromotion);
```
ericrkuo, line 157: Whats an inputtedPromotion? Maybe promotionRequestBody

src/test/controller/PromotionController.test.ts
```diff
@@ -259,13 +264,14 @@ describe('Unit tests for PromotionController', function () {
   test('POST /promotions/ - if restaurant with placeId does not exist in DB, promotion should create new restaurant', async (done) => {
     const user: User = new UserFactory().generate();
     const promotion = new PromotionFactory().generateWithRelatedEntities(user);
+    const inputtedPromotion = setGooglePlacesAddress(promotion);
```
ericrkuo, line 267: I think we can write some more tests
- one where restaurantAddress is empty string
- another where no restaurant exists and and we create restaurant with geocoder and check with the expected lat lon
- one where geocoder returns no results and for now, we create restaurant with null lat and null lon

src/test/service/GeocodingService.test.ts
```diff
@@ -0,0 +1,57 @@
+import { GeocodingService } from '../../main/service/GeocodingService';
+import { GeocodingObject } from '../../main/data/GeocodingObject';
+import { randomString } from '../utility/Utility';
+
+describe('tests for Geocoding Service', function () {
+  let geocoder: GeocodingService;
+
+  beforeAll(() => {
+    geocoder = new GeocodingService();
```
ericrkuo, line 9: I think it's not that good to use API keys for tests,  That's why I used a mock for the Google place service test. 

I think there was a free provider that didn't require an api key, thing it was called open maps or something. We could make geocodingservice constructor take in the configuration  for node geocoder, or make it take in a node geocoder instance too

Another thing we could do is see if we can find a mocking library for node geocoder. Last resort would be to make the mock class ourself!

package.json
```diff
@@ -30,16 +30,17 @@
     "@googlemaps/google-maps-services-js": "3.1.16",
     "@types/bcrypt": "^3.0.0",
     "@types/bcryptjs": "^2.4.2",
+    "@types/node-geocoder": "^3.24.1",
     "@types/redis": "^2.8.28",
     "@types/redis-mock": "^0.17.0",
     "@types/request-promise": "^4.1.47",
-    "@firebase/app-types": "0.6.1",
```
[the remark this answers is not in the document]
ld357: We are! It's weird cause the top has `"@firebase/app-types": "0.x"` but I'll have to double check. When I ran `yarn install` it updated the version so I undid the change, but I'll probably add it back in again

src/main/controller/PromotionController.ts
```diff
@@ -103,23 +111,37 @@ export class PromotionController {
...
+          }
+          await transactionalEntityManager
+            .getCustomRepository(RestaurantRepository)
+            .save(restaurant);
```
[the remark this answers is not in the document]
ld357: I thought we need to because we're not creating the restaurant and saving it anywhere else in the code?

src/main/controller/PromotionController.ts
ld357: Are you suggesting a new name for the function or is this the name of the function for the abstraction?

src/main/data/GeocodingObject.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Used with GeocodingService for a restaurant's lat/lon values
+ */
+export interface GeocodingObject {
```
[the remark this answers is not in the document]
ld357: Hmm, geocoordinate means 1 coordinate though?

src/main/service/GeocodingService.ts
ld357: I just mentioned it, is it bad to include it?

src/main/service/GeocodingService.ts
ld357: I couldn't find any documentation on whether or not it was the most accurate so I just decided to get the first one

src/main/service/GeocodingService.ts
ld357: Ah I forgot to put this in the description but I tested out 3 different locations with 3 providers and compared the results to what Google Places API would give us. Then I averaged out the differences and locationiq had the lowest of 0.01880393333.

src/main/service/GeocodingService.ts
ld357: Yup, https://locationiq.com/pricing

There's no billing required during signup and the free tier accepts 5000 quests/day

src/test/controller/PromotionController.test.ts
ld357: I put input as in that's the input promotion we're sending in, in contrast to expected/output/result

src/test/service/GeocodingService.test.ts
ld357: I considered the mock you had but I think it's fine considering that we're in the free tier, there's no storage of info and no chance we'll get charged for over usage

package.json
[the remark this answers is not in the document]
ericrkuo: Oh lets use this one, and not "0.x" then

src/main/controller/PromotionController.ts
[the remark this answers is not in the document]
ericrkuo: I believe we are saving restaurant. Notice how we aren't calling discount repository to save the promotions discount and or schedule repository to save the schedules

src/main/controller/PromotionController.ts
ericrkuo: What's the difference?

src/main/data/GeocodingObject.ts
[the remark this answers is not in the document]
ericrkuo: Not sure! This was first word that came to my mind, looks like .NET uses similar terminology https://docs.microsoft.com/en-us/dotnet/api/system.device.location.geocoordinate?view=netframework-4.8

src/main/service/GeocodingService.ts
ericrkuo: Oh I'm just not too sure what it means for a service class to be a Node library

src/main/service/GeocodingService.ts
ericrkuo: hmm let's see if we can spend some time trying to figure out what's the most accurate! There's different ways we can approach this
- one idea we could do is make the request body take in google places lat lon, and compare the lat lon values with those ones and see which ones the closest. But I'll leave it to you to explore and try out which you think is best! I think taking the first one is not the best idea unless its the most accurate

src/main/service/GeocodingService.ts
ericrkuo: Oh I see, was the free one, OpenMaps or something, not good? We may want to test out more than 3 locations.

src/test/service/GeocodingService.test.ts
ericrkuo: Hmm let's picture it this way:

the free tier has 60 requests per minute
Currently, in our test class, we make 5 requests
Let's say theoretically we have 6 PRs currently up and running integration tests at the same time - 30 requests left
At the same time 2 developers are developing locally and each test the geocoding 10 times - 10 requests left
And 100 customers are uploading promotions (sadly no more requests left until we wait another minute)

Similar reasons can be seen for the 5000 request limit
- https://softwareengineering.stackexchange.com/questions/184979/unit-testing-open-source-projects-that-require-keys-for-api-calls
- https://softwareengineering.stackexchange.com/questions/214154/how-can-i-unit-test-a-class-which-requires-a-web-service-call/214159#:~:text=Your%20unit%20test%20should%20not,you%20want%20fast%20unit%20tests.

src/main/controller/PromotionController.ts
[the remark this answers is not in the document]
andrewzulaybar: Ah is this because of the `cascade: true`? https://github.com/ubclaunchpad/PromoPal-backend/pull/123#discussion_r589914723

src/main/service/GeocodingService.ts
andrewzulaybar: Oh yeah maybe just make it more clear that it uses a Node library under the hood? I think rn it just sounds like it is a Node library

src/main/service/GeocodingService.ts
andrewzulaybar: Could you add it to `sample.env` and also add the key in the Google Drive folder?

src/main/service/GeocodingService.ts
andrewzulaybar: Just did a quick search on this and it looks like there are a few options: https://stackoverflow.com/questions/18883601/function-to-calculate-distance-between-two-coordinates

tl;dr either use the Haversine formula (I think we used this in d3 for cpsc 310) or google maps also has a function for it, but that includes installing it in the BE too https://developers.google.com/maps/documentation/javascript/reference#spherical

src/main/validation/PromotionValidation.ts
```diff
@@ -43,6 +43,7 @@ export class PromotionValidation {
       .default(new Date()),
     /** note incoming format is a string, and Joi will automatically convert to type Date */
     expirationDate: Joi.date().required(),
+    googlePlacesAddress: Joi.string().required(),
```
andrewzulaybar, line 46: How about `address`? That way, we're not coupling it to the implementation details / choice of library

src/test/service/GeocodingService.test.ts
andrewzulaybar: Ahh this is a good point. If we can mock it out, I think that would be better practice

andrewzulaybar (approved):
This looks good! Really glad you guys brought this idea up :) 

Just left some more comments, mainly piggy-backing off of Eric's 😛

src/main/controller/PromotionController.ts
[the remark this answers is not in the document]
ericrkuo: Yup exactly!

src/main/service/GeocodingService.ts
ericrkuo: Yeahh for this approach we would need the request body to take in the lat lon from google places, I'm okay with using the first tbh, I think we can refactor and consider a more complex approach once we see an actual bug where the lat/lon is completely off

src/main/service/GeocodingService.ts
ericrkuo: Looked into locationiq, think its a good provider! Let's stick with this one

ericrkuo (approved):
Great work! Really appreciate how much work you put into this. I think if we can figure out a way for our tests to not use the API key, then we can merge it in!

src/main/controller/PromotionController.ts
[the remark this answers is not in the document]
ld357: Ooooh, I see!! I never saw that comment oops but I'll remove :)

src/main/service/GeocodingService.ts
ld357: Thanks to the both of you for researching more into it 😄  I can take a deeper look into it this week! If it takes too long based on timeline, since we're trying to merge everything by this weekend, I might just stick with first one but will keep you posted via Slack!

src/main/service/GeocodingService.ts
```diff
@@ -0,0 +1,33 @@
...
+  public async getGeoCoordinateFromAddress(
+    address: string
+  ): Promise<GeoCoordinate> {
+    const entries: Entry[] = await this.geocoder.geocode(address);
+    const result: GeoCoordinate = {};
+
+    if (entries?.length) {
+      const geocodeResult = entries[0];
+      result.lat = geocodeResult.latitude;
+      result.lon = geocodeResult.longitude;
```
ericrkuo, line 28: Maybe create a ticket and put a small blurb explaining why we're taking the first in the array?

src/test/controller/BaseController.ts
```diff
@@ -64,3 +67,9 @@ export const createFirebaseMock = (): Auth => {
   );
   return mockSdk.auth();
 };
+
+export const createGeocodingServiceMock = (): GeocodingService => {
+  return new GeocodingService({
+    provider: 'openstreetmap',
+  });
+};
```
ericrkuo, line 75: Excellent!!!!

src/test/service/GeocodingService.test.ts
```diff
@@ -0,0 +1,47 @@
...
+import { randomString } from '../utility/Utility';
+
+describe('tests for Geocoding Service', function () {
+  let geocoder: GeocodingService;
+
+  beforeAll(() => {
+    geocoder = new GeocodingService({
+      provider: 'openstreetmap',
+    });
+  });
```
ericrkuo, line 12:
```suggestion
    geocoder = createGeoCodingServiceMock();
```

src/main/App.ts
```diff
@@ -50,10 +51,16 @@ export class App {
 
       this.redisClient = await this.createRedisClient();
       this.firebaseAdmin = await initFirebaseAdmin();
+      const geocodingService = new GeocodingService({
+        provider: 'locationiq',
+        apiKey: process.env.GEOCODING_KEY,
+      });
+
```
ericrkuo, line 58: Maybe rather than passing in the actual service class, we can pass in the node geocoder instance!

```suggestion
      const nodeGeoCoder = node_geocoder({
        provider: 'locationiq',
        apiKey: process.env.GEOCODING_KEY,
      })
```

Then `GeocodingService's` constructor will take in a parameter of type `Geocoder` rather than of type `Options`. 
And then also `createGeocodingServiceMock` could be `createMockNodeGeocoder` which returns type `Geocoder`.

src/test/service/GeocodingService.test.ts
ericrkuo: Or actually `geocodingService = new GeoCodingService(createMockGeoCoder())`

ericrkuo (approved):

src/main/App.ts
ericrkuo: The reason is say in the future imagine we have another service class that needs `node-geocoder`, we would need to add in more parameters.

ericrkuo: And also like `createGeocodingServiceMock` suggests that we are mocking the service class, but actually, we are mocking `node-geocoder`, our service class isn't being mocked.

ericrkuo:
Great job @ld357! Just some minor comments about the mocking, but other than that great work!

ericrkuo:
@ld357 also don't forget to change the base branch of your PR to master! I forgot to change my other PRs as well

src/main/App.ts
ld357: Oooh yeah those are good points!! Sure I can update them :)

src/main/App.ts
```diff
@@ -89,7 +97,7 @@ export class App {
     const restaurantRouter = new RestaurantRouter(restaurantController);
     app.use(Route.RESTAURANTS, restaurantRouter.getRoutes());
 
-    const promotionController = new PromotionController();
+    const promotionController = new PromotionController(nodeGeocoder);
```
ericrkuo, line 100: Oh wait sorry one minor thing, could we pass in the GeocodingService class into PromotionController? I noticed that we're instantiating a geoCodingService inside PromotionController. The slight problem with that is we are no longer doing dependency injection since PromotionController is now saying "I will instantiate a PromotionService" rather than, "I have been injected a PromotionService".

here's an example of how we do it for restaurants
image

ld357: Oooh yeah sure I can change it, I thought you wanted me to pass in the instance instead of the actual service class earlier but I'll change it back!

andrewzulaybar (approved):

PR #126, opened 2021-03-11
Pp 28 52 save delete promotion

ericrkuo:
BE portion for: https://github.com/ubclaunchpad/PromoPal-frontend/pull/25

Problem:
We want to know when calling `GET /promotions` whether a user saved a promotion or not.

Approach:
- Added a non-column property to promotion `isSavedByUser`
- `GET /promotions` now can take in a `userId` param.
   - We could authenticate this endpoint and use the user from the authentication, but we may not want this endpoint to be protected
- PromotionRepository has a new method to find promotions saved by the user and populate the field `isSavedByUser`
- Added tests


src/main/repository/PromotionRepository.ts
```diff
@@ -16,7 +22,11 @@ export class PromotionRepository extends Repository<Promotion> {
...
       JSON.stringify(promotionQuery) !== JSON.stringify({})
     ) {
-      return this.applyQueryOptions(promotionQuery);
+      const promotions = await this.applyQueryOptions(promotionQuery);
+      if (promotionQuery.userId && promotions.length != 0) {
+        return this.findPromotionsUserSaved(promotionQuery.userId, promotions);
+      }
+      return promotions;
     } else {
       return this.createQueryBuilder('promotion')
```
ericrkuo, line 31: This class is getting ugly IMO and too complex. The next time we add more stuff to here, I'll do some refactoring and try to abstract it more

andrewzulaybar (approved):
Just tested it out locally on the branch that I'm working on and it looks good to me! Thanks for getting this up so quick :)

src/main/repository/PromotionRepository.ts
```diff
@@ -16,7 +22,11 @@ export class PromotionRepository extends Repository<Promotion> {
       promotionQuery &&
       JSON.stringify(promotionQuery) !== JSON.stringify({})
     ) {
-      return this.applyQueryOptions(promotionQuery);
+      const promotions = await this.applyQueryOptions(promotionQuery);
+      if (promotionQuery.userId && promotions.length != 0) {
```
ld357, line 26: Just a minor thing but I think we can condense more!

```suggestion
      if (promotionQuery.userId && promotions.length) {
```

src/test/validation/PromotionQueryValidation.test.ts
```diff
@@ -275,6 +276,18 @@ describe('Unit tests for PromotionQueryValidation', function () {
     }
   });
 
+  test('Should fail if userId is not uuid', async () => {
+    try {
+      promotionQueryDTO.dayOfWeek = Day.MONDAY;
```
ld357, line 281: Did we mean to change this line to promotionQueryDTO.userId equals some random non-uuid?

ld357 (approved):
Two small comments, otherwise looks good to me! Going to approve :)

src/test/validation/PromotionQueryValidation.test.ts
ericrkuo: Ohya good point, that's strange I have no idea why the test passed. I'll take a look thanks!

src/main/repository/PromotionRepository.ts
ericrkuo: Good point thanks!

src/test/validation/PromotionQueryValidation.test.ts
ericrkuo: Ah I realized it's because I didn't insert the `fail("Should have failed")` right after calling the validation

PR #127, opened 2021-03-15
PP-72, PP-73, PP-89: Add ability to sort promotions

andrewzulaybar:
**Description:**
Originally in the FE (https://github.com/ubclaunchpad/PromoPal-frontend/pull/20), we decided to move sorting to the BE because we should ideally be sorting before we send the promotions to the FE

**Changes:**
- Added `install-extensions.sh` script to install `cube` and `earthdistance` postgres extensions to for calculating distance. I've also added this to the docker-compose file and I'm not entirely sure if this is the right place to put this, but I couldn't get it to work otherwise... I also think to get it working locally you need to delete your local `tmp` folder before running `docker-compose up`.
- Added support for the `sort` query param (also on the FE) so that the queries are expected to be of the following forms:

1. /sort=Distance&lat=49.12&lon=-123.201
2. /sort=Popularity
3. /sort=Recency

**Related PRs:**
https://github.com/ubclaunchpad/PromoPal-frontend/pull/32

**Related tickets:**
https://promopal.atlassian.net/browse/PP-72
https://promopal.atlassian.net/browse/PP-73
https://promopal.atlassian.net/browse/PP-89


install-extensions.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" -d foodies <<-EOSQL
+CREATE EXTENSION cube;
+CREATE EXTENSION earthdistance;
+EOSQL
```
ericrkuo, line 8: Neat! What are extensions in PostgreSQL?

src/main/data/Sort.ts
```diff
@@ -0,0 +1,5 @@
+export enum Sort {
+  DISTANCE = 'Distance',
+  POPULARITY = 'Popularity',
+  RECENCY = 'Recency',
+}
```
ericrkuo, line 5: Maybe `SortOptions` and some documentation for what this means?

ericrkuo (commented):
Looks good! I'm just gonna take some time to checkout your branch and test this out, curious to see how it works! I've never heard of extensions, super cool

install-extensions.sh
andrewzulaybar: I think they're kind of like modules/libraries that you can import? Tbh didn't know about them until I looked into this task 😛 

From this [blog post](https://severalnines.com/database-blog/creating-new-modules-using-postgresql-create-extension#:~:text=PostgreSQL%20extension%20functionality%20allows%20for,function%20as%20built%2Din%20features.):
> PostgreSQL database has many features and offers a wide range of data types, functions, operators etc. But sometimes it is not enough for some use cases. We can easily extend PostgreSQL's functionality through extensions. This is very useful feature for database developers and administrators.

And in this case it allows us to calculate distances between lat/lon values with simpler syntax (and I think the alternative would be to use some equation like the harversine formula)

src/main/data/Sort.ts
andrewzulaybar: Oooh yes I can add this!

install-extensions.sh
ericrkuo, line 8:
```
CREATE EXTENSION IF NOT EXISTS cube;
CREATE EXTENSION IF NOT EXISTS earthdistance;
```

I think I have an idea of how we can make this work without having to create a shell script. We can actually create a migration. Take a look at https://typeorm.io/#/migrations, and our repo also has some migrations too.

Basically, the migration would be to run the script above. I've configured migrations to run when our application starts. So we wouldn't need to worry about docker-compose and since we have the `IF NOT EXISTS`, we don't need to worry about installing it each time we start our application.

src/main/repository/PromotionRepository.ts
```diff
@@ -117,6 +118,10 @@ export class PromotionRepository extends Repository<Promotion> {
       return this.fullTextSearch(queryBuilder, promotionQuery);
     }
 
+    if (promotionQuery?.sort) {
```
ericrkuo, line 121: I think we should also check that lat/lon are defined

src/main/validation/PromotionQueryValidation.ts
```diff
@@ -45,5 +49,8 @@ export interface PromotionQueryDTO {
   cuisine?: CuisineType | CuisineType[];
   expirationDate?: Date;
   dayOfWeek?: Day;
+  sort?: SortOptions;
+  lat?: number;
+  lon?: number;
```
ericrkuo, line 54: Kind of a challenge if you want, totally okay if you don't want to:

How about we add some validation such that if the sort option is Distance, then we need to make sure lat/lon are defined

ericrkuo (approved):

ericrkuo:
@andrewzulaybar Changes mainly from this commit https://github.com/ubclaunchpad/PromoPal-backend/pull/127/commits/bebf321511acc649d56a33deb292a14ed5ec4382, let me know if there's any way to improve it, but I tested it out locally and it seems to work

ericrkuo:
If you're merging, let's make sure to uncomment https://github.com/ubclaunchpad/PromoPal-frontend/pull/36/files#diff-a63e6281f7980eaec57fcd34c6c303bf906188e866bb29330d01c2c8a3e28d46R122-R133 since it should work now

src/main/validation/PromotionQueryValidation.ts
```diff
@@ -27,6 +28,17 @@ export class PromotionQueryValidation {
     /** note incoming format is a string, and Joi will automatically convert to type Date */
     expirationDate: Joi.date(),
     dayOfWeek: Joi.string().valid(...Object.values(Day)),
+    sort: Joi.string().valid(...Object.values(SortOptions)),
+    lat: Joi.any().when('sort', {
+      is: SortOptions.DISTANCE,
+      then: Joi.number().required(),
+      otherwise: Joi.number(),
+    }),
```
ericrkuo, line 36: I found out Joi has this cool "if then" syntax. I've made it so that if we've specified sort by distance, then we need to make sure lat and lon are both defined

andrewzulaybar: Oooh this is actually really cool, glad you found this!!!

richardyjtian (approved):

PR #128, opened 2021-03-21
Voting backend implementation

yunishimura0716:
I just attempt to implement the voting system on Backend side.

I created new Entity, `VoteRecord` as intermediate table, which contains `user` and `promotion` and VoteState (INIT, UP, DOWN).

The flow of up voting is as follow:
- The user request POST `/promotions/:id/upVote` with request body {uid: user id}
- In Backend, try to find VoteRecord using `uid` and `pid`.
- If fail, create new VoteRecord
- If the state of VoteRecord is `UP`, return (no change)
- Otherwise, increment the `votes` of the promotion.
- After that, if the state of VoteRecord is `INIT`, the next state is `UP`. if the state is `DOWN`, the next state is `INIT`.

down vote is also similar flow (contrast way).

I'm welcome any comment or improvement from your suggestion.


src/main/entity/VoteRecord.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  @ManyToOne(() => User, (user) => user.votedRecord, {
+    onDelete: 'CASCADE',
+  })
+  user: User;
+
+  @ManyToOne(() => Promotion, (promotion) => promotion.votedBy, {
+    onDelete: 'CASCADE',
+  })
+  promotion: Promotion;
```
ericrkuo, line 38: So just to double-check the cascade deletes are working

- [ ] if we delete promotion, the respective vote records get deleted?
- [ ] if we delete user, the respective vote records get deleted?

ericrkuo (commented):
@yunishimura0716 I thought we were going to wait for design team (elina + wendy) to see if we were going to replace upvote/downvote with the hearts (saved promotions) instead.

So rather than seeing the number of upvotes/downvotes like reddit, instead we would see the number of hearts (number of saved promotions). Could you check with @andrewzulaybar or design team if this is the case?

src/main/entity/VoteRecord.ts
yunishimura0716: I added the test to check the cascade works.

ericrkuo: Awesome thanks! Great job by the way, this is a very detailed PR.

yunishimura0716:
I added that every time the user request get promotions, each has the state of vote (`INIT` or `DOWN` or `UP`). 

Also added the test for that in `EntityIntegration.test`.

src/main/controller/PromotionController.ts
```diff
@@ -183,12 +185,50 @@ export class PromotionController {
...
+        } else {
+          // user vote first time
+          await transactionalEntityManager
+            .getCustomRepository(PromotionRepository)
+            .increment({ id: pid }, 'votes', 1);
+        }
+        voteRecord.voteState =
+          voteRecord.voteState === VoteState.INIT
+            ? VoteState.UP
+            : VoteState.INIT;
```
ericrkuo, line 228: I was thinking something along the lines of this for how it would flow:
image

- For example in case 2, if the user has already downvoted the promotion, they can either:
    - upvote it which will highlight the up arrow 
    - click downvote again to remove their vote (which is setting the vote state to INIT)

Because of this, I was thinking we would need three endpoints
`/upvote` 
`/downvote`
`/resetVote` 

So I was thinking `/upvote` endpoint would follow this logic:
```
1. Find if a vote record exists
    1.  If the vote state is DOWN, set it to UP and increase votes by 2
    2.  If already UP, don't do anything and return
    3. if INIT set it to UP and increase votes by 1
2. If vote record does not exist
    1. Call repository.save({ userId: uid, promotionId: pid, voteState: VoteState.UP})
    2. Increase the number of votes by 1
```

This is an example of reddit where you can see they increase votes by 2 when I jump from down to up
RedditVoteSystem

yunishimura0716: Thank you for your instruction, it is so much helpful to implement the voting flow. I'm just curious why we need `resetVote`??

ericrkuo: I see, yah actually we could make the upvote endpoint set the state to INIT if the vote record exists and the state was UP.

I was thinking of having it as a separate endpoint like 'resetVote' because if we make a request to 'upvote', I don't really expect the vote to go to INIT. Then it isn't really clear that the upvote endpoint is actually upvoting.

It's totally up to you though, this is a design decision. Actually, one thing I'd recommend is to see how reddit is doing it by inspecting the networks tab and seeing what they send in the request body or if it's 2/3 different endpoints or all one endpoint

yunishimura0716: Ok, so I gonna follow the logic you gave (how reddit is doing) and then let you know after that!

yunishimura0716: I saw how reddit is doing and it seems to use the same endpoint for up and down vote. It provides the form data with the request which includes `id` and `dir` where `dir` might have `1` , `0` or `-1`. If the status becomes `UP` (INIT => UP or DOWN => UP), the dir is `1`. If the status becomes `INIT` (UP => INIT or DOWN => INIT), the dir is `0`. If the status becomes `DOWN`, the dir is `-1`.

ericrkuo: Very neat! Hmm I see, so I'm guessing that for this case:

> If the status becomes UP (INIT => UP or DOWN => UP), the dir is 1

If it's INIT => UP, then we add 1 to total votes
If it's DOWN => UP, then we add 2 to total votes
So at the controller level, we would need to know the previous VoteState (if it exists)

I see, either way, I think we can be very flexible with how we approach it and I'll let you choose which one you think is the best!

yunishimura0716:
I modified the flow of up and down vote.
The flow of up vote is as follow:
```
1. Find if a vote record exists
    1.  If the vote state is DOWN, set it to UP and increase votes by 2
    2.  If already UP, set it to INIT and decrease votes by 1
    3. if INIT set it to UP and increase votes by 1
2. If vote record does not exist
    1. Call repository.save({ userId: uid, promotionId: pid, voteState: VoteState.UP})
    2. Increase the number of votes by 1
```

src/main/controller/PromotionController.ts
```diff
@@ -183,12 +185,54 @@ export class PromotionController {
...
+            .getCustomRepository(VoteRecordRepository)
+            .create({ userId: uid, promotionId: pid, voteState: VoteState.UP });
+          await transactionalEntityManager
+            .getCustomRepository(VoteRecordRepository)
+            .save(voteRecord);
+          await transactionalEntityManager
+            .getCustomRepository(PromotionRepository)
+            .increment({ id: pid }, 'votes', 1);
+          return response.status(204).send();
+        }
```
ericrkuo, line 218: How about this? I think it's a bit easier to read

```suggestion
        let voteRecord = await transactionalEntityManager
            .getCustomRepository(VoteRecordRepository)
            .findOne({ userId: uid, promotionId: pid });;
        if (!voteRecord) {
            await createVoteRecord(uid, pid, VoteState.UP, transactionalEntityManager)
            return response.status(204).send
        }
```

and actually, we can combine
```
          voteRecord = await transactionalEntityManager
            .getCustomRepository(VoteRecordRepository)
            .create({ userId: uid, promotionId: pid, voteState: VoteState.UP });
          await transactionalEntityManager
            .getCustomRepository(VoteRecordRepository)
            .save(voteRecord);

into this:

          await transactionalEntityManager
            .getCustomRepository(VoteRecordRepository)
            .save({ userId: uid, promotionId: pid, voteState: VoteState.UP }));

```

src/main/entity/Promotion.ts
```diff
@@ -14,6 +14,7 @@ import { PromotionType } from '../data/PromotionType';
 import { CuisineType } from '../data/CuisineType';
 import { SavedPromotion } from './SavedPromotion';
 import { Schedule } from './Schedule';
+import { VoteRecord, VoteState } from './VoteRecord';
```
ericrkuo, line 17: Just some linting errors, I usually run `yarn lintFix` before I push and it'll console output any warnings

src/main/entity/Promotion.ts
```diff
@@ -179,9 +180,18 @@ export class Promotion {
   })
   votes: number;
 
+  @OneToMany(() => VoteRecord, (voteRecord) => voteRecord.promotion, {})
+  votedBy: VoteRecord[];
```
ericrkuo, line 184: Some documentation for this?

src/main/repository/PromotionRepository.ts
```diff
@@ -216,6 +224,43 @@ export class PromotionRepository extends Repository<Promotion> {
...
+  ): Promise<Promotion[]> {
+    const promotionIds = promotions.map((promotion: Promotion) => promotion.id);
+    const voteRecordIds = await getConnection()
+      .createQueryBuilder(VoteRecord, 'voteRecords')
+      .select(['voteRecords.promotionId', 'voteRecords.voteState'])
+      .where('voteRecords.userId = :userId', { userId })
+      .andWhere('voteRecords.promotionId IN (:...promotionIds)', {
+        promotionIds,
+      })
+      .getRawMany();
```
ericrkuo, line 245: Ah sorry I made changes in my other PR for `findPromotionUserSaved`. 

Try following this, the benefit is that it will return an array of VoteRecord, so we can leverage type safety at compile time
https://github.com/ubclaunchpad/PromoPal-backend/pull/131/files#diff-45ba25f3f1fe17fcdfb9215942531a829664bc5de9e9bec00c3e08a1e25ff66aL203-R212

src/main/entity/Promotion.ts
```diff
@@ -179,9 +180,18 @@ export class Promotion {
...
    * True if the user has saved this promotion
    * * Note this is used for endpoints that need to be aware of the user making the request
    * */
   isSavedByUser?: boolean;
+
+  /**
+   * True if the user has voted this promotion
+   * This is used for endpoints that need to be aware of the user making the request
+   */
+  voteState?: number;
```
ericrkuo, line 196: Does this work?
```suggestion

  /**
   * Vote state of promotion
   * This is used for endpoints that need to be aware of the user making the request
   */
  voteState?: VoteState;
```

ericrkuo (commented):
Looks good! Just some minor comments

src/main/controller/PromotionController.ts
yunishimura0716: Thanks Eric! What is `createVoteRecord` function ? Did you make?

ericrkuo: Oh no I didn't make it, but I meant like we can pull it out into its own method

yunishimura0716: I see, thanks! I'll make clean the piece of code more!

src/main/repository/PromotionRepository.ts
yunishimura0716: Oh, ok. I'll refer it!

src/main/repository/PromotionRepository.ts
```diff
@@ -216,6 +225,41 @@ export class PromotionRepository extends Repository<Promotion> {
...
+      })
+    );
+    return promotions.map((promotion: Promotion) => {
+      if (dict.has(promotion.id)) {
+        promotion.voteState = dict.get(promotion.id);
+      } else {
+        promotion.voteState = VoteState.INIT;
+      }
+      return promotion;
+    });
```
ericrkuo, line 261: I think we can simplify the code, how do you feel about this?

```suggestion
    const promotionIdToVoteState = new Map<string, VoteState>(
      voteRecords.map((voteRecord) => {
        return [voteRecord.promotionId, voteRecord.voteState];
      })
    );
    return promotions.map((promotion: Promotion) => {
      promotion.voteState = promotionIdToVoteState.get(promotion.id) ?? VoteState.INIT
      return promotion;
    });
```

src/test/repository/EntityIntegration.test.ts
```diff
@@ -928,4 +932,60 @@ describe('Integration tests for all entities', function () {
...
+        Number(
+          promotions.find((promotion) => promotion.id === promotion2.id)
+            ?.voteState
+        )
+      ).toEqual(VoteState.DOWN);
+      expect(
+        Number(
+          promotions.find((promotion) => promotion.id === promotion3.id)
+            ?.voteState
+        )
```
ericrkuo, line 985: I don't think we need the `Number(...)` part, we should be able to compare using `voteState`

ericrkuo (approved):
Excellent work @yunishimura0716, really appreciate how much time and effort you put into this. This is great!

src/main/repository/PromotionRepository.ts
yunishimura0716: Oh, thanks for great simplified code!
I'll do that!

src/test/repository/EntityIntegration.test.ts
yunishimura0716: The reason why I put Number is there's weird error that the voteState becomes `string` not number when compare. But , there's no such error anymore, so yeah I gonna remove Number.

ericrkuo:
Looks good to me! Seems like you'll need one more reviewer before you can merge

ld357 (commented):
This looks great, Yu!! Awesome job implementing this!! Just left a couple small refactoring comments, otherwise looks good to me 👍🏻

ld357 (approved):
Thanks for making those changes!! 🥳  Looks good 👍🏻

PR #129, opened 2021-03-26
PP-88 Get all promotions from a restaurant

ericrkuo:
https://promopal.atlassian.net/browse/PP-81

We want to be able to get all promotions for a specific restuarant


andrewzulaybar (approved):
This looks great, thanks! Tested it out on the FE and it works :)

ld357 (approved):
Nice!! 🥳  Tested on postman and works 🙌🏻

PR #130, opened 2021-03-28
move ts-node and tsc from devDep -> dependencies

na-wu:


ericrkuo:
I think `pg, query-string, @types/express, @types/node, body-parser, joi, typeorm`  will all need to be in "dependencies" section rather than "dev-dependencies" as well, could we move those?

`@types/redis-mock, firebase-mock, redis-mock` should all be in "dev-dependencies" rather than "dependencies" 

I think we can also delete these dependencies, they aren't being used
https://github.com/ubclaunchpad/PromoPal-backend/blob/c2b903fc2869de01228ea06885ebf840f3d6822a/package.json#L31-L32

ericrkuo:
Thanks for catching these bugs haha

ericrkuo (approved):

andrewzulaybar:
In addition to what Eric said I think you might need to run `yarn install` again after moving the dependencies around to update the `yarn.lock` file?

ericrkuo:
@na-wu I updated the rest of the dependencies, should be good to merge, just need one more approval from @andrewzulaybar

ericrkuo:
> In addition to what Eric said I think you might need to run `yarn install` again after moving the dependencies around to update the `yarn.lock` file?

It did not change when I ran it, strange!

andrewzulaybar (approved):

PR #131, opened 2021-03-30
PP-85 implement deleting images

ericrkuo:
https://promopal.atlassian.net/browse/PP-85
FE PR: https://github.com/ubclaunchpad/PromoPal-frontend/pull/40

Problem:
We want to delete promotions images when we delete a promotion or delete a user (which deletes the users uploaded promotions)

Approach:
- Created `ResourceCleanupService` to handle external resource cleanup with respect to a promotion
    - this class talks to our S3 bucket https://s3.console.aws.amazon.com/s3/buckets/promopal?region=us-west-1&tab=objects
- Used https://www.npmjs.com/package/mock-aws-s3 for tests
    - The neat thing about this mock is it actually writes the `Objects` as files into our application. I've put the files in the .gitIgnore

Additional Stuff:
- Refactored `BaseController` into a class and made the mocks properties of the class. That way reduced duplicate code in our integration test files and makes it easier to remember what we are mocking and where the mocking is


sample.env
```diff
@@ -6,4 +6,6 @@ FIREBASE_CLIENT_EMAIL=INSERT_HERE
 FIREBASE_WEB_API_KEY=INSERT_HERE
 FIREBASE_SERVICE_ACCOUNT_ID=INSERT_HERE
 GOOGLE_API_KEY=your_key
-GEOCODING_KEY=INSERT_HERE
\ No newline at end of file
+GEOCODING_KEY=INSERT_HERE
+S3_ACCESS_KEY_ID=INSERT_HERE
+S3_SECRETE_ACCESS_KEY=INSERT_HERE
```
ld357, line 11: Small typo on SECRET!

src/main/service/ResourceCleanupService.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * Cleanup any external resources associated with multiple promotions
+   * * E.g. Delete S3 images associated with many promotions
+   * @param promotionIds the id of the promotions to clean up resources for
+   * */
+  cleanupResourceForPromotions(
+    promotionIds: string[]
+  ): Promise<PromiseResult<DeleteObjectsOutput, AWSError>> {
```
ld357, line 46: Just curious 😄  but how does it work if there's an AWSError?

src/test/controller/BaseController.ts
```diff
@@ -5,71 +5,97 @@ import { auth } from 'firebase-admin/lib/auth';
 import Auth = auth.Auth;
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const firebaseMock = require('firebase-mock');
-import { AxiosInstance } from 'axios';
+import axios, { AxiosInstance } from 'axios';
 import nodeGeocoder, { Geocoder } from 'node-geocoder';
+import AWSMock from 'mock-aws-s3';
+import { S3 } from 'aws-sdk';
+import { S3_BUCKET } from '../../main/service/ResourceCleanupService';
```
ld357, line 12: I can't remember if we have other constants like this, but maybe potentially adding a config file if we have more?

src/test/controller/BaseController.ts
```diff
@@ -5,71 +5,97 @@ import { auth } from 'firebase-admin/lib/auth';
...
-  return mockSdk.auth();
-};
+  /**
+   * Creates an express app and registers all handlers and routes
+   * * Used for integration tests that test controllers (supertest needs an instance of express app without the port binding)
+   * */
+  registerTestApplication = async (): Promise<Express> => {
+    const app = new App();
+    const expressApp = express();
+    await app.registerHandlersAndRoutes(
```
ld357, line 36: Should we wrap these in try/catch?

src/test/controller/BaseController.ts
```diff
@@ -5,71 +5,97 @@ import { auth } from 'firebase-admin/lib/auth';
...
-  await app.registerHandlersAndRoutes(
-    expressApp,
-    redisClient,
-    firebaseAdmin,
-    geocoder,
-    axiosInstance
-  );
-  return expressApp;
-};
+export class BaseController {
```
ld357, line 14: This is super clean and smart! Thanks for refactoring it nicely 🥳

src/test/controller/BaseController.ts
```diff
@@ -5,71 +5,97 @@ import { auth } from 'firebase-admin/lib/auth';
...
+    await this.mockS3.deleteBucket({ Bucket: S3_BUCKET }).promise();
+    return expressApp;
+  };
 
-export const createMockNodeGeocoder = (): Geocoder => {
-  return nodeGeocoder({
-    provider: 'openstreetmap',
-  });
-};
+  static createRedisMock = (): RedisClient => {
```
ld357, line 49: Maybe specify these to `public`?

yarn.lock
```diff
@@ -295,7 +295,7 @@
     minimatch "^3.0.4"
     strip-json-comments "^3.1.1"
 
-"@firebase/app-types@0.6.1", "@firebase/app-types@0.x":
```
ld357, line 298: Thanks for fixing this!!

ld357 (approved):
Great job with the cleanup service and also abstracting out the BaseController!! This looks super difficult so kudos 🎉   I left a couple of comments for discussion!

sample.env
ericrkuo: Ah man my spelling is getting worse and worse, thanks!

src/main/service/ResourceCleanupService.ts
```diff
@@ -0,0 +1,59 @@
...
+   * */
+  cleanupResourceForPromotion(
+    promotionId: string
+  ): Promise<PromiseResult<DeleteObjectOutput, AWSError>> {
+    const deleteObjectRequest: DeleteObjectRequest = {
+      Bucket: S3_BUCKET,
+      Key: promotionId,
+    };
+
+    return this.s3.deleteObject(deleteObjectRequest).promise();
```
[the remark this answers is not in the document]
ericrkuo: Ohh I see, using folders within buckets, this is a great idea for the future once we start supporting more and more images. Right now we only have one image per promotion. I think we should definitely do this once we start expanding, do you want to create a ticket for this?

> Also just out of curiosity- does this completely wipe everything with the key, regardless of file type?

Yeah the key must be unique, so it will only delete the single object with the requested key, regardless of file type

src/test/controller/BaseController.ts
ericrkuo: I think it's okay, not catching exceptions in tests is okay I think because we are leaving it to the testing framework to output that an unexpected error occured

src/test/controller/BaseController.ts
ericrkuo: Oh the default should be `public` right?

src/main/service/ResourceCleanupService.ts
ericrkuo: Good Q! not sure, I will need to look into this hehe

ericrkuo: Okie after looking at it, I think it works like this: 

`<PromiseResult<DeleteObjectsOutput, AWSError>` is defined as this:
image
So actually, we are returned
`DeleteObjectsOutput & {$response: Response<DeleteObjectsOutput. AWSError>}`

Then we can go into the individual type definitions of `DeleteObjectsOutput, Response, AWSError` to see how the data will be handed to us.

src/test/controller/BaseController.ts
ericrkuo: I think it should be okay since this is one constant and its in a service class. It's usage is only in the service class asides from test classes using it

yarn.lock
ericrkuo: Haha this was by accident, didnt know i made this change

src/main/service/ResourceCleanupService.ts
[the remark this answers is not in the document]
ericrkuo: So I looked into this some more, S3 does have a concept of folders, it automatically makes them folders if they share the same relative path, so basically the keys need to be paths themselves
https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-folders.html

Based on my quick research, there isn't an easy way to delete all objects in a folder, but I did see some workarounds where you can basically find all keys that start with the root folder. So for example if we grouped promotion images by userId and had the following keys
```
user1/promotion1ID
user1/promotion2ID
user1/promotion3ID
```
we would find all objects prefixed with user1, and delete those objects. In other words, we've deleted all resources associated with user1.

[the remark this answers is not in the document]
ld357: Ooo yeah!! That makes sense, I can make a ticket and include the research you've done here :) I've ran into a similliar thing during work, and we use variables in our config file to maintain the keys that form the path too!

src/main/service/ResourceCleanupService.ts
ld357: Cool!! Thanks for looking into it 👍🏻  Should we check into this object wherever we're making the call to the function?

src/test/controller/BaseController.ts
ld357: Hmm I think you're right, I just suggested it cause we've noted the visibility for our other files too so not sure if we want to stay consistent!

ericrkuo: Ah yeah we have a mix of methods specifying public or not, I usually don't but I agree we should stay consistent. I think the majority do not specify public, these are the only ones I could find. But when it comes to `private/protected` we need to specify the visibility

image

src/main/service/ResourceCleanupService.ts
ericrkuo: Yeahh I was thinking about that, usually when I see an error its thrown, I haven't seen an instance of `AWSError`. I don't think we need to worry about it, our use case of this is very small just deleting an object regardless if it exists or not.

yunishimura0716 (approved):
Thanks for great work, @ericrkuo . The piece of code makes sense to me. The things I was wondering become clear by Lily.

sample.env
```diff
@@ -6,4 +6,6 @@ FIREBASE_CLIENT_EMAIL=INSERT_HERE
 FIREBASE_WEB_API_KEY=INSERT_HERE
 FIREBASE_SERVICE_ACCOUNT_ID=INSERT_HERE
 GOOGLE_API_KEY=your_key
-GEOCODING_KEY=INSERT_HERE
\ No newline at end of file
+GEOCODING_KEY=INSERT_HERE
+S3_ACCESS_KEY_ID=INSERT_HERE
+S3_SECRET_ACCESS_KEY=INSERT_HERE
```
ericrkuo: @ubclaunchpad/foodies I've updated the env file with new environment variables https://drive.google.com/drive/u/0/folders/1YhX35Vk8I0_PECrk6I2btsFPdd8BC-ow. Please take a look as I put a comment how to get your S3 env variables

PR #132, opened 2021-04-01
PP-98 DELETE user endpoint needs to be protected

ld357:
`DELETE /users/:id` needs to be:

**1. Protected by firebase auth to ensure the user is authenticated to make the request**
=> Yu already accomplished this functionality by adding `this.firebaseAuth.isAuthorizedForProtection` to the route!

**2. Check the correct user is making the request**
=> A user does not have permission to delete another user's account

I also created an Error class to handle these cases and future cases!


src/main/middleware/FirebaseAuth.ts
```diff
@@ -26,6 +26,7 @@ export class FirebaseAuth {
       const decodedToken = await this.admin.verifyIdToken(idToken);
 
       if (decodedToken) {
+        res.locals.firebaseUserId = decodedToken.uid;
```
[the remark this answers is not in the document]
ericrkuo: This is a great idea to get the firebase id in the middleware! 

We used like `app.locals` in Flow which is bad because that's for the entire application. I'm not sure of the impact between `req.locals` or `res.locals`

But how about this alternative? I'm kind of scared with using `locals`.
```suggestion
        req.headers.firebaseUserId = decodedToken.uid;
```

ericrkuo:
Ohya for point 1, we just need to add Yu's middleware to the delete router! Similar to 
https://github.com/ubclaunchpad/PromoPal-backend/blob/bc94c9f1e17189b367b201bf82ef1fb056c2f95f/src/main/route/UserRouter.ts#L50-L53

src/main/controller/UserController.ts
```diff
@@ -162,6 +162,20 @@ export class UserController {
...
+          {
+            cache: true,
+          }
+        );
+
+        // todo: handle null authenticatedUser case
+
+        if (authenticatedUser?.id !== id) {
+          // todo: throw error
+        }
```
ericrkuo, line 177: Yup, exactly the idea we were having in mind, great job! We don't need the cache since the user _should_ be deleted.

ericrkuo:
Great work 🎉

src/main/middleware/FirebaseAuth.ts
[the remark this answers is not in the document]
ld357: Yeah! I think we can use it for the deleting promotion scenario too 🥳   

Oooh right!! I never heard about `req.headers` but I just tried to do some quick research except I couldn't find anything :( I'm down to change though if its commonly used in this way!

I referenced this link when I was implementing it and seems like it should be safe? ([Stack Overflow - Passing Variables to the next middleware using next in expressJS](https://stackoverflow.com/questions/18875292/passing-variables-to-the-next-middleware-using-next-in-express-js))

ld357:
> Ohya for point 1, you're right that Yu implemented the middleware function, but we need to specify the middleware in the router so that when we make the call to the endpoint, it'll get called! Similar to
> https://github.com/ubclaunchpad/PromoPal-backend/blob/bc94c9f1e17189b367b201bf82ef1fb056c2f95f/src/main/route/UserRouter.ts#L50-L53

Oooh for this- we have it specified in the UserRouter already!

ericrkuo:
> > Ohya for point 1, you're right that Yu implemented the middleware function, but we need to specify the middleware in the router so that when we make the call to the endpoint, it'll get called! Similar to
> > https://github.com/ubclaunchpad/PromoPal-backend/blob/bc94c9f1e17189b367b201bf82ef1fb056c2f95f/src/main/route/UserRouter.ts#L50-L53
> 
> Oooh for this- we have it specified in the UserRouter already! 

Ohcrap sorry was thinking of the wrong router, looks good then!

src/main/middleware/FirebaseAuth.ts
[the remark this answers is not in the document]
ericrkuo: Ohh I see, okay res.locals makes more sense now! Thanks for the link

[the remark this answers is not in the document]
ericrkuo: Oh 'req.headers' is just the headers of a request, I was thinking of just adding a new header, but after taking a look at the link, res.locals is the recommended approach by express

src/main/middleware/ErrorHandler.ts
```diff
@@ -14,18 +14,19 @@ export function errorHandler(
   next: NextFunction
 ): any {
   // todo: see if can figure out the extension functions
-  if (error instanceof QueryFailedError)
```
ld357, line 17: I removed this case because its the same as the default!

yunishimura0716 (approved):
Thanks for great work, @ld357 !! The piece of code is readable and good additional test for delete endpoint.

src/main/errors/Error.ts
```diff
@@ -0,0 +1,18 @@
+class BaseError extends Error {
+  public name: string;
+  public status: number;
+
+  constructor(errorName: string, errorMessage: string, status: number) {
+    super(errorMessage);
+    this.name = errorName;
+    this.status = status;
+  }
+}
```
ericrkuo, line 10: Ooh I like this abstraction, nice job!

ericrkuo: What's the reason behind separating the two files, `libraries/Error.ts` and `errors/Error.ts`?

src/main/middleware/ErrorHandler.ts
```diff
@@ -14,18 +14,19 @@ export function errorHandler(
...
     );
+
+  let statusCode = 400;
+
+  if (error instanceof ForbiddenError) statusCode = 403;
+
   return res
-    .status(400)
+    .status(statusCode)
     .send(new FrontEndErrorObject(error.name, [error.message]));
```
ericrkuo, line 31: How about this? Now we make it generic to the type contract of BaseError. Any other error is out of our control and maybe should be status 500 b/c its an unexpected internal server error
```
  if (error instanceof BaseError)
    return res.status(error.status).send(
        new FrontEndErrorObject(
            error.name,
            [error.message]
        )
    )

  return res
    .status(500)
    .send(new FrontEndErrorObject(error.name, [error.message]));
```

src/test/controller/UserController.test.ts
```diff
@@ -222,9 +222,57 @@ describe('Unit tests for UserController', function () {
       });
   });
 
-  test('DELETE /users/:id', async (done) => {
+  test('DELETE /users/:id - ForbiddenError should be caught', async (done) => {
```
ericrkuo, line 225: Maybe we can make the title less technical like:
```suggestion
  test('DELETE /users/:id - Should not be able to delete another user', async (done) => {
```

src/test/controller/UserController.test.ts
```diff
@@ -222,9 +222,57 @@ describe('Unit tests for UserController', function () {
...
+      .end((err, res) => {
+        const frontEndErrorObject = res.body;
+        expect(frontEndErrorObject?.errorCode).toEqual('ForbiddenError');
+        expect(frontEndErrorObject.message).toHaveLength(1);
+        expect(frontEndErrorObject.message[0]).toEqual(
+          'Your account does not have sufficient privileges to perform this action.'
+        );
+        done();
+      });
+  });
```
ericrkuo, line 248: I was thinking this might make it easier for the developer to understand which user is which. What do you think?

```suggestion
    const otherUser: User = new UserFactory().generate();
    otherUser.firebaseId = 'randomfirebaseId';

    const userWhoWantsToDeleteOtherUser: User = new UserFactory().generate();
    userWhoWantsToDeleteOtherUser.firebaseId = firebaseId;

    await userRepository.save(otherUser);
    await userRepository.save(userWhoWantsToDeleteOtherUser);

   // simulating user who is making request to delete another use
    request(app)
      .delete(`/users/${otherUser.id}`)
      .set('Authorization', idToken)
      .expect(403)
      .end((err, res) => {
        const frontEndErrorObject = res.body;
        expect(frontEndErrorObject?.errorCode).toEqual('ForbiddenError');
        expect(frontEndErrorObject.message).toHaveLength(1);
        expect(frontEndErrorObject.message[0]).toEqual(
          'Your account does not have sufficient privileges to perform this action.'
        );
        done();
      });
  });
```

src/main/middleware/ErrorHandler.ts
ld357: Ooo!! I like this :) will update, did you want me to add back in QueryFailedError for status 400 or have it as 500 as well?

src/test/controller/UserController.test.ts
ld357: Hmmm, how about `deletingUser` and `authenticatingUser`?

src/main/errors/Error.ts
ld357: I wanted `libraries/Error.ts` to be more readable and we can customize the message (or other details in the future)- so it's kind of like a helper/factory file that uses errors from the `errors/Error.ts` class!

src/main/middleware/ErrorHandler.ts
ericrkuo: I think 500 as well sounds good to me

src/test/controller/UserController.test.ts
ericrkuo: Whichever you think makes it more clear!

src/main/errors/Error.ts
ericrkuo: Hmm I see, I think devs may be confused whether to use the Error.ts from libraries or the one from errors. For me I'm not too sure which one to use when looking at it.

Imo I like the idea of the `errors/Error.ts` class, I find that's sufficient enough. What we could do is have an `errors/ErrorMessages.ts` and do something like:

```
class ForbiddenError extends BaseError {
  constructor(errorMessage?: string) {
    super('ForbiddenError', errorMessage ?? ErrorMessages.INSUFFICIENT_PRIVELEGES, 403);
  }
}

// then we can just call
throw new ForbiddenError();
```

The benefit of this is that all errors are contained in one file and more simplicity since there's less confusion between libraries/errors directory.

ld357: Hmm yeah I can see how you can get confused! I'll update to something similar to what you suggested here!

src/test/controller/UserController.test.ts
```diff
@@ -286,29 +286,14 @@ describe('Unit tests for UserController', function () {
...
-              userRepository.findOneOrFail({ id: expectedUser.id })
+              userRepository.findOneOrFail({ id: userToDelete.id })
             ).rejects.toThrowError();
             done();
           }
         );
       });
   });
 
-  test('DELETE /users/:id - delete non-existent user should not fail', async (done) => {
```
ld357, line 249: I removed this test because it's no longer valid or possible to test now that we have the auth protection! This is because we have two scenarios:

1. Non-existent user is the authenticated user
2. Non-existent user is not the authenticated user

If 1 is the case, then we'll either save or not save the user before making the call to the endpoint. If we save, then the user is no longer nonexistent, but if we don't save, it'll throw an "EntityNotFound" when we check for the user using the firebaseId which is not what we're trying to test here (already tested)!

If 2 is the case, then a "ForbiddenError" will be thrown, which again is not what we're trying to test here (already tested)!

src/main/repository/UserRepository.ts
```diff
@@ -11,7 +11,7 @@ export class UserRepository extends Repository<User> {
 
   findByFirebaseId(
     firebaseId: string,
-    options: FindOneOptions<User>
+    options?: FindOneOptions<User>
   ): Promise<User | undefined> {
```
ericrkuo, line 15: Oo I realized we can do this!
```suggestion
    options?: FindOneOptions<User>
  ): Promise<User> {
```

src/main/controller/UserController.ts
```diff
@@ -162,6 +163,15 @@ export class UserController {
...
           UserRepository
         );
+
+        const authenticatedUser = await userRepository.findByFirebaseId(
+          res.locals.firebaseUserId
+        );
+
+        if (authenticatedUser?.id !== id) {
+          throw new ForbiddenError();
+        }
```
ericrkuo, line 173: We should be able to do this now with my other suggestion!
```suggestion
        if (authenticatedUser.id !== id) {
          throw new ForbiddenError();
        }
```

src/test/controller/UserController.test.ts
ericrkuo: Excellent points, thanks for explaining this! Great job for looking into other parts of our code that are affected with your changes like this test case

src/main/middleware/ErrorHandler.ts
```diff
@@ -14,18 +14,20 @@ export function errorHandler(
...
+
+  if (error instanceof BaseError)
+    return res
+      .status(error.status)
+      .send(new FrontEndErrorObject(error.name, [error.message]));
+
   return res
-    .status(400)
+    .status(500)
     .send(new FrontEndErrorObject(error.name, [error.message]));
```
ericrkuo, line 32: 😍

src/main/middleware/ErrorHandler.ts
```diff
@@ -1,7 +1,7 @@
 import { NextFunction, Request, Response } from 'express';
-import { QueryFailedError } from 'typeorm';
 import { FrontEndErrorObject } from '../data/FrontEndErrorObject';
 import { ValidationError } from 'joi';
+import { BaseError, ForbiddenError } from '../errors/Error';
```
ericrkuo, line 4: Small lint warning!

src/main/errors/Error.ts
```diff
@@ -0,0 +1,24 @@
...
+
+class ForbiddenError extends BaseError {
+  constructor(errorMessage?: string) {
+    super(
+      'ForbiddenError',
```
ericrkuo, line 17: Minor thing, but what do you think about an `ErrorCodes` class, so we could do something like `ErrorCodes.FORBIDDEN_ERROR`, similar to `ErrorMessages`. Totally okay if you think its too much

ericrkuo (approved):
Fix it then ship it as my coworkers say! 🚢

src/main/errors/Error.ts
ld357: Hmm, I think it's okay for now but I think we could def consider it for the future when we have more error types and whatnot! :)

PR #133, opened 2021-04-04
PP-97 delete promotion needs to be protected and check correct user

ericrkuo:
https://promopal.atlassian.net/browse/PP-97
Very similar to https://github.com/ubclaunchpad/PromoPal-backend/pull/132 except we want to make sure a user can only delete a promotion they uploaded (or in other words a user cannot delete any other uses promotions)


src/main/middleware/FirebaseAuthMiddleware.ts
```diff
@@ -6,7 +6,7 @@ import Auth = auth.Auth;
 /**
  * Middleware function to check if the request has correct client header using firebase-admin auth
  */
-export class FirebaseAuth {
+export class FirebaseAuthMiddleware {
```
ericrkuo, line 9: Renamed for more clarity because I got confused what the difference was between `FirebaseAuth` and `Auth`

src/main/route/UserRouter.ts
```diff
@@ -1,71 +1,72 @@
...
+  private firebaseAuthMiddleware: FirebaseAuthMiddleware;
 
-  constructor(userController: UserController, firebaseAdmin: Auth) {
+  constructor(
+    userController: UserController,
+    firebaseAuthMiddleware: FirebaseAuthMiddleware
+  ) {
     this.userController = userController;
-    this.firebaseAuth = new FirebaseAuth(firebaseAdmin);
+    this.firebaseAuthMiddleware = firebaseAuthMiddleware;
```
ericrkuo, line 15: Fixed the dependency injection here!

src/test/controller/BaseController.ts
```diff
@@ -43,6 +45,8 @@ export class BaseController {
     );
     // cleanup resources from previous bucket if possible
     await this.mockS3.deleteBucket({ Bucket: S3_BUCKET }).promise();
+    await (this.mockFirebaseAdmin as any).autoFlush();
+    await this.createAuthenticatedUser();
```
ericrkuo, line 49: Pulling this up to simplify all our integration controller tests

src/main/controller/UserController.ts
```diff
@@ -155,7 +155,10 @@ export class UserController {
     }
   };
 
-  // delete one user
+  /**
+   * Delete a user.
+   * We need to ensure that a user can only delete their own account.
```
ld357, line 160: Thanks for adding this in! :)

ld357 (approved):
Awesome job cleaning up everything, looks simpler to understand now!! 🤩  💯

src/test/controller/BaseController.ts
```diff
@@ -94,6 +98,19 @@ export class BaseController {
...
   };
 
+  /**
+   * Create a mock authenticated user
+   * */
+  async createAuthenticatedUser(): Promise<void> {
+    // create user
+    const user = await (this.mockFirebaseAdmin as any).createUser({
+      email: 'test@gmail.com',
+      password: 'testpassword',
```
yunishimura0716, line 108: I'm just wondering the part `(this.mockFirebaseAdmin as any)` . What does this do?

src/main/controller/PromotionController.ts
```diff
@@ -168,11 +170,24 @@ export class PromotionController {
...
         });
-        const promotion = await transactionalEntityManager
+
+        const authenticatedUser = await transactionalEntityManager
+          .getCustomRepository(UserRepository)
+          .findByFirebaseId(response.locals.firebaseUserId);
+
+        const potentialPromotionUploadedByUser = await transactionalEntityManager
+          .getCustomRepository(PromotionRepository)
+          .findOne({ id, user: authenticatedUser });
```
yunishimura0716, line 180: @ericrkuo , this is good way to find authorized user.  Could you change user deletion in the same way to make the implementation consistent between promotion and user?

yunishimura0716 (commented):
Thanks great work, @ericrkuo ! Overall, our BE functionality is refactored and getting improved as OOD!

src/test/controller/BaseController.ts
ericrkuo: Ohya good question, before we had it like [this](https://github.com/ubclaunchpad/PromoPal-backend/blame/bc94c9f1e17189b367b201bf82ef1fb056c2f95f/src/test/controller/UserController.test.ts#L48-L51), and `mockFirebaseAdmin` was already type `any`. I've just casted it back to `any` because our mocking library seems to complain if it isn't `any`.  I think this is because the mock-firebase-admin doesnt have supported typescript, but let me double check if I can fix this

src/main/controller/PromotionController.ts
ericrkuo: Ohh I see, kk will do good point

yunishimura0716 (approved):
Thanks Eric! Nice work!

PR #134, opened 2021-04-08
PP-108 Register user to firebase in BE

ericrkuo:
https://promopal.atlassian.net/browse/PP-108

What this PR covers
- We want to create a user through firebase in our BE. This was mainly to avoid the complex flow of creating a firebase user in FE and then creating the user in our database by sending POST req to BE
    - This means we must send email and password for our `UserDTO` when making a POST req
    - Created `FirebaseService` to handle creation of a user in firebase
- We are no longer storing `User::email` in our database. Firebase will store our email for us.
   - Pro: when we need to update email of our user, only need to update it in firebase, not our DB.
- **We are no longer storing `User::firebaseId`**
   - When we create a user in firebase, we will configure so that the id of the user in our DB is the same as the firebaseId in firebase
   - Pro: Less concern of mixing around `User::firebaseId` and `User::id`
   - I read online that leaking the firebaseId is totally okay, because user's still don't have access to the idToken, password, or email. See this excellent explanation: https://stackoverflow.com/questions/37221760/firebase-is-auth-uid-a-shared-secret
- Also changed `GET UploadedPromos` to return array of promotions, otherwise FE is broken


src/main/service/FirebaseService.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Service class for operations related to firebase
+ */
+export class FirebaseService {
+  private admin: Auth;
+
+  constructor(firebaseAdmin: Auth) {
+    this.admin = firebaseAdmin;
+  }
```
ericrkuo, line 12: Benefit of dependency injection: Notice how our controller tests don't need to go about a heavy mess to make sure that `FirebaseService::firebaseAdmin` is not the real `Auth` we will be using in production, instead it will be a mock.

src/test/controller/BaseController.ts
```diff
@@ -99,16 +104,30 @@ export class BaseController {
...
+    this.authenticatedUser = authenticatedUser;
+  }
+
+  /**
+   * Delete the mock authenticated user
+   * * Note this should be called in the afterEach of test files (we do not want tests to rely on the same user in multiple tests)
+   * */
+  async deleteAuthenticatedUser(): Promise<void> {
+    await this.mockFirebaseAdmin.deleteUser(this.authenticatedUser.id);
   }
```
ericrkuo, line 131: I don't like how future devs working on controller tests will need to be responsible for rememberingg to specify these two methods in `beforeEach` and `afterEach`. Will think about this some more...

src/test/controller/UserController.test.ts
```diff
@@ -82,57 +86,57 @@ describe('Unit tests for UserController', function () {
...
+        expect(user.firstName).toEqual(userDTO.firstName);
+        expect(user.lastName).toEqual(userDTO.lastName);
+        expect(user.username).toEqual(userDTO.username);
+
+        // firebaseId should be the same as the userId
+        const firebaseUser = await baseController.mockFirebaseAdmin.getUser(
+          user.id
+        );
+        expect(firebaseUser).toBeDefined();
+        expect(firebaseUser.email).toEqual(userDTO.email);
```
ericrkuo, line 113: Important test for POST /users

src/test/controller/BaseController.ts
```diff
@@ -85,7 +88,9 @@ export class BaseController {
       // }
       null
     );
-    return mockSdk.auth();
+    const firebaseAdmin = mockSdk.auth();
+    (firebaseAdmin as any).autoFlush();
+    return firebaseAdmin;
```
yunishimura0716, line 93: Why did you make this change?

yunishimura0716 (commented):
Thank you for your great refactoring!! The piece of codes makes sense to me.
Is that all you need to do? or Do you have any sections you haven't implemented yet??

src/test/controller/BaseController.ts
ericrkuo: Ohya i was thinking its too much for developers to remember to call auto Flush for tests when they use mock firebase admin. In my FirebaseService.ts class, notice how I only need to call mockFirebaseAdmin() without needing to remember to call autoFlush. Originally, I forgot to call autoFlush and my tests were timing out. 

Would this be okay?

ericrkuo:
> Thank you for your great refactoring!! The piece of codes makes sense to me.
> Is that all you need to do? or Do you have any sections you haven't implemented yet??

Yeah I think this is all I need to do. One thing to note now is that when we verify the idToken in our middleware, res.locals.firebaseUserId will be the same id as the one in our database

src/test/controller/BaseController.ts
yunishimura0716: Oh I see, thanks!

yunishimura0716 (approved):

src/main/controller/UserController.ts
```diff
@@ -119,8 +100,18 @@ export class UserController {
...
-        return res.status(201).send({ ...result, firebaseId: undefined });
+        const savedUser = await userRepository.save(user);
+        try {
+          await this.firebaseService.createUserFromEmailAndPassword(
+            savedUser.id,
+            userDTO.email,
+            userDTO.password
+          );
+        } catch (e) {
+          await userRepository.delete(savedUser);
```
richardyjtian, line 111: Would it be better to register in firebase first then DB, and then delete from firebase if it fails (like the original FE design)? I think it's more likely that firebase will return an error from registering with an existing email.

ericrkuo: Oh crap sorry I did not see this.

Hmm I see, so then in our controller we would need to generate a random uid, pass it to firebase, and make sure the user saved has that same uid.

I think I like the idea of deleting our user in DB upon failure rather than deleting our user in firebase. Deleting user in DB should be much faster than deleting user in firebase, although I do agree that registering with an existing email would probably have more errors.

src/main/controller/UserController.ts
```diff
@@ -337,12 +324,19 @@ export class UserController {
         const userRepository = transactionalEntityManager.getCustomRepository(
```
andrewzulaybar, line 324: Is this still needed 👀

andrewzulaybar (approved):
Haven't tested it out locally yet but will soon :) LGTM

src/main/controller/UserController.ts
ericrkuo: Good catch thanks!

src/main/controller/UserController.ts
richardyjtian: Sure yeah either works!

PR #136, opened 2021-04-10
Improved geocoder results by adding in country code

ld357:
Set the country code of our geocoder query to Canada, through a new `GeocoderConfig` object, which will drastically improve our geocoding results!

[View the sample locations and results tested here](https://docs.google.com/document/d/1I4-WOaSAlEtvmGyNcjKy36puAmFO9qq_wpdjhMwiTOc/edit?usp=sharing)


src/main/App.ts
```diff
@@ -53,10 +53,16 @@ export class App {
...
+
       const geocoder = nodeGeocoder({
         provider: 'locationiq',
         apiKey: process.env.GEOCODING_KEY,
       });
+      const countryCode = 'ca';
+      const geocoderConfig: GeocoderConfig = {
+        geocoder: geocoder,
+        countryCode: countryCode,
+      };
```
yunishimura0716, line 65: Do you think it is ok to hard code `ca` for `countryCode` variable? I mean, for now, every time we change the country code, other developer have to find this area. If we won't change the value in the future, that's fine!

yunishimura0716 (commented):
Thanks your work, @ld357 !
It make the code more readable (good variable name etc) !

src/main/App.ts
ld357: Ooo that's an excellent point! I'm not too sure if we'll change this variable in the future, but I can put a comment above this line or in the function itself to denote the default is currently Canada?

ericrkuo: Does country code need to be 'Ca'?

ericrkuo, line 65:
```suggestion
      const geocoderConfig: GeocoderConfig = {
        geocoder: geocoder,
        countryCode: 'ca',
      };
```

src/main/service/GeocodingService.ts
```diff
@@ -1,14 +1,20 @@
-import { Entry, Geocoder } from 'node-geocoder';
+import { Entry, Geocoder, Query } from 'node-geocoder';
 import { GeoCoordinate } from '../data/GeoCoordinate';
 
+export interface GeocoderConfig {
+  geocoder: Geocoder;
+  countryCode?: string;
+}
```
ericrkuo, line 7: Maybe a comment here why we need country code? and in what scenario its undefined

src/test/controller/BaseController.ts
```diff
@@ -28,7 +29,10 @@ export class BaseController {
   constructor() {
     this.mockRedisClient = BaseController.createRedisMock();
     this.mockFirebaseAdmin = BaseController.createFirebaseMock();
-    this.mockGeoCoder = BaseController.createMockNodeGeocoder();
+    const mockGeocoder = BaseController.createMockNodeGeocoder();
+    this.mockGeocoderConfig = {
+      geocoder: mockGeocoder,
+    };
```
ericrkuo, line 35: I was thinking something like this to stay more consitent with what we have
```suggestion
    this.mockGeocoderConfig = BaseController.createMockNodeGeocoderConfig();
```

src/test/service/GeocodingService.test.ts
```diff
@@ -7,9 +7,11 @@ describe('tests for Geocoding Service', function () {
...
 
   beforeAll(() => {
-    geocodingService = new GeocodingService(
-      BaseController.createMockNodeGeocoder()
-    );
+    const mockGeocoder = BaseController.createMockNodeGeocoder();
+    const geocoderConfig = {
+      geocoder: mockGeocoder,
+    };
+    geocodingService = new GeocodingService(geocoderConfig);
```
ericrkuo, line 14: And then this can be more succinct
```suggestion
    geocodingService = new GeocodingService(
      BaseController.createMockNodeGeocoderConfig()
    );
```

ericrkuo (approved):

ericrkuo:
Great work @ld357, can't wait to show off the new location accuracy tmr!

src/main/App.ts
ld357: I think we're focusing on Canada for now? @andrewzulaybar could you clarify? :)

ericrkuo: Yeah just Canada for now. @yunishimura0716 that's a good point, and we definitely want our geoCoder class to support multiple country codes. I think this is something to definitely think about when we make our app work outside of Canada.

yunishimura0716 (approved):
Thank you for your refactoring! Good job!

src/main/App.ts
andrewzulaybar: Yeah Canada for now!

andrewzulaybar (approved):
Looks great!

